PR #3, opened 2019-10-09
Feature: implement button for GitHub and Stack Overflow

kaicataldo:
I took a crack at implementing this! To get this working, I downloaded and added the standalone build and the parsers from `https://prettier.io/lib/*.js`. This obviously isn't ideal, but I figured the team would know best how to go about getting the standalone build in the extension!

Some other notes:
1. `ext/_locales/en/messages.json` is necessary because a `default_locale` is listed in `manifest.json`. So far, this is a very easy thing to translate (there's only one string to display!), and it seems like it would be worthwhile.
1. GitHub was easy because comments are all markdown! 
1. Stack Overflow is trickier because it uses a different syntax to denote the language of code snippets.
1. It would be great to have some kind of icon that fits the Stack Overflow design better, but I think this is a good MVP!

For reference, here's what it currently looks like:

Stack Overflow:
Screen Shot 2019-10-09 at 1 28 48 AM

GitHub:
Screen Shot 2019-10-09 at 1 28 58 AM


ext/src/content/content.js
```diff
@@ -0,0 +1,232 @@
...
+        });
+        target.focus();
+      });
+    }
+
+    containerEl.addEventListener("select", handleGitHubTextareaEvents);
+    containerEl.addEventListener("keyup", handleGitHubTextareaEvents);
+  }
+
+  function initStackOverflowBtn() {
```
lipis: `initStackOverflowButton`

ext/src/content/content.js
```diff
@@ -0,0 +1,232 @@
...
+    for (const [key, value] of Object.entries(style)) {
+      button.style[key] = value;
+    }
+
+    append ? el.append(button) : el.prepend(button);
+
+    return button;
+  }
+
+  function initGitHubBtn() {
```
lipis: `initGitHubButton`

LOLdevelopr:
Great. This extension is really going to help a lot in these website.

.eslintrc.json
```diff
@@ -1,58 +1,65 @@
...
   "env": {
-    "browser": true,
     "node": true
   },
   "plugins": ["prettier"],
   "extends": ["prettier"],
   "parserOptions": {
     "ecmaVersion": 2020,
-    "sourceType": "module",
+    "sourceType": "script",
```
kaicataldo: We're not currently using modules, so I think this is safer?

.eslintrc.json
```diff
@@ -1,58 +1,65 @@
...
-    "strict": [2, "global"]
-  }
+    "sort-vars": "error",
+    "strict": ["error", "global"]
+  },
+  "overrides": [
+    {
+      "files": ["ext/**/*"],
+      "env": {
+        "browser": true
```
kaicataldo: Scoping the browser env to just the files that are executed by the browser

.eslintrc.json
```diff
@@ -1,58 +1,65 @@
...
   "rules": {
-    "curly": 2,
-    "dot-notation": 2,
-    "id-length": 2,
-    "no-const-assign": 2,
-    "no-dupe-class-members": 2,
-    "no-else-return": 2,
-    "no-inner-declarations": 2,
-    "no-lonely-if": 2,
+    "curly": "error",
```
kaicataldo: I'm a maintainer of ESLint, and we generally recommend using `"error"` instead of numbers for clarity!

kaicataldo:
Sorry, diff got a little weird with the renamed files after the rebase. I realized that this doesn't implement the button for the GitHub Issues view, but I think it would be nice to land this and follow up with that if the team is amenable to that.

lipis:
So what works now?! (Will be able to test in few)

kaicataldo:
Should work on StackOverflow and GitHub Pull Requests!

kaicataldo:
I'd like to figure out a better way to detect the when to render the buttons for the GitHub view (since editable comments can be created on the fly) so that we can render it immediately (instead of waiting until the user inputs text), but I think this is a good MVP to build on.

lipis:
It didn't work for me.. On SO I can see the button, but it doesn't seem to do anything, and can't see it on GitHub..

kaicataldo:
Hmmm strange - here's what I'm seeing:

lipis:
are you sure you committed all the files?

kaicataldo:
Ah, I'm seeing it go in and out on GitHub - sometimes it works and sometimes it doesn't. I assumed it was all prerendered because it feels blazing fast to me and I was seeing it work locally all the time, but it's most likely that sometimes the element it's binding to isn't on the page when it runs. Will probably have to do something similar to the SO one and poll for the element to be in the document.

lipis:
It's a great start for sure 😊

kaicataldo:
For reference, here's what it looks like on GitHub for me:

kaicataldo:
Can you see if that fixes it for you? The Prettier button should appear when you start editing a comment. I think we can definitely come up with a better check that makes the button appear when the comment appears (would be a nicer UX), but I thought this was a good first pass to get the basic functionality :)

lipis:
The button disappears when answering anonymously on SO

lipis:
So the button works (on SO) when there is code in the answer.. otherwise it doesn't formats the markdown (for example fixing the links and lists, spaces).

kaicataldo:
> So the button works (on SO) when there is code in the answer.. otherwise it doesn't formats the markdown (for example fixing the links and lists, spaces).

This is fixed.

lipis:
I think it's a pretty solid start.. Merging it if you don't have anything to add at this point.

kaicataldo:
Sounds good! Definitely lots of room for improvement :)

lipis:
Do you mind adding issues with things needs to be done or improved, optimized whatever you think of so we won't forget anything.

kaicataldo:
Already on it :smile:

kaicataldo:
Also, thanks for the quick reviews! ❤️

Issue #12, opened 2019-10-09
Fix button rendering behavior for logged out commenting on Stack Overflow


kaicataldo:
The button currently disappears after it is rendered because Stack Overflow renders additional information:
<img width="633" alt="Screen Shot 2019-10-09 at 3 23 53 PM" src="https://user-images.githubusercontent.com/7041728/66513322-d6c2b280-eaa8-11e9-9ebe-0d11b90b8037.png">

Issue #15, opened 2019-10-09
Fix: figure out a better way to render the GitHub comment buttons


kaicataldo:
The GitHub buttons currently wait until a user starts typing before rendering. It would be great to figure out some way to know when a new comment is created (maybe an `onclick` listener, filtering out by what element was clicked and checking the existence of a new comment?). We should be able to traverse the DOM for to find the relevant `textarea` when the Prettier button is clicked, so it should be able to be a fairly naive check just to scan for any editable textareas that don't have a button.

nickmccurdy:
Fixed by #19

Issue #14, opened 2019-10-09
Fix: add Prettier button for issues (new and existing)


kaicataldo:

nickmccurdy:
#19 works for me when commenting and editing existing issues, but not when creating new issues.

Wicked7000:
@nickmccurdy I can see the button when creating issues - Do you still see this as I can't seem to find a case on my side where it is not working.

nickmccurdy:
I tried it again in this repo and it seems to be working fine now. Can be closed, I'll make another issue if I find a reproducible regression.

kaicataldo:
LGTM!
<img width="1040" alt="Screen Shot 2019-10-14 at 10 39 30 PM" src="https://user-images.githubusercontent.com/7041728/66795737-81850780-eed3-11e9-9409-326d2a2dd7d8.png">

PR #19, opened 2019-10-10
Use tree walking and polling for Github buttons

Wicked7000:
This pull request is to address the following issues:
#15 
#14 

What does this change do?
- Uses tree walking to find text-area's 
  - First we find all the buttons we want to create 'Prettier' buttons next to and create them. Once we have done that we then do a tree walk from that button upwards until we find the first text area.
- Added buttons to Issues

I've tested this in most cases I think of with no issues found.
- New PR's
- New Issues
- Issue comments
- PR comments (as well as comments on diffs)

The only thing that might want tweaking is the Polling timer I've setup, I've set it to a second as realistically the check that I'm doing to see if the path name has changed is extremely light-weight but we could change it to something like five seconds if you don't mind the buttons appearing slightly later.

### DIFF
--- a/ext/src/content/content.js
+++ b/ext/src/content/content.js
@@ -2,7 +2,7 @@
 
 function init() {
   const GITHUB_URL = "https://github.com";
-  const GITHUB_VALID_PATHNAMES = /^\/.*\/.*\/(?:pull\/\d+(?:\/?|\/files\/?)$|commit|issues\/\d+|issues\/new)/u;
+  const GITHUB_VALID_PATHNAMES = /^\/.*\/.*\/(?:pull\/\d+(?:\/?|\/files\/?)$|commit|compare\/.*|issues\/\d+|issues\/new)/u;
   const GITHUB_POLLING_TIME_IN_SECONDS = 1;
   let isGithubListenerAdded = false;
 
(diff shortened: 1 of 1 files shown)


nickmccurdy:
The code looks good and it seems to work pretty well.

I'm not sure we need the timeout though, could we wait for some DOM event like load or DOMContentLoaded instead? It should be more reliable when the page takes a longer time than expected to load.

kaicataldo:
Thanks for doing this! I agree with @nickmccurdy. Is there some kind of event we could listen for? Or could we just bind a listener to clicking on any buttons that trigger an editable comment to appear?

I'm not seeing it work for a few cases:

### Adding a comment in the PR diff:

Screen Shot 2019-10-10 at 11 40 19 PM

### Editing an existing comment in the PR conversation view:

Screen Shot 2019-10-10 at 11 39 51 PM

Finally, from a UX perspective, I personally don't think we should move any of the existing GitHub buttons around. I know I click `Cancel` enough that I might accidentally click `Prettier` without thinking:

Screen Shot 2019-10-10 at 11 39 57 PM

kaicataldo:
Maybe we could use [`MutationObserver`](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver)? I _think_ that like as long as we wait until after the first render, we should be able to create an observer on each `HTMLElement` that the comment renders into?

Wicked7000:
Yeah they seem like fair observations, I'll see if I can switch over to MutationObserver (I also think the buttons not showing is probably caused by the timeouts)

Wicked7000:
I've changed over to using mutationObserver which seems to make things more consistent and got rid of the polling behaviour, let me know if you have any other suggestions.

Also the button will always now appear as the furthest left button (so that it does not get confused with the Cancel button or close pull request buttons from a UX point of view as mentioned)

lipis:
The Prettier button doesn't appear when I'm trying to edit a comment (for example in this PR)

Wicked7000:
> The Prettier button doesn't appear when I'm trying to edit a comment (for example in this PR)
This should be fixed with the latest commit, I have to add additional observers for the comment area's (that have to be setup again on page-change) but this seems to work well and is reasonably fast as the comment subtrees are fairly small.

Apologies for not checking it I forgot to check github comments.

Wicked7000:
test

Wicked7000:
Also needed an extra observer to deal with new comments (otherwise the listener won't be there for when they click edit) -> again this is only checking immediate sub-tree so it should be fine.

kaicataldo:
Functionality-wise, this looks like it's working great for me! It looks like some of the styling could be further refined, but I would advocate for landing this and fixing those cases up afterwards. This is a much better solution than the one I came up with in the initial implementation ❤️

kaicataldo:
Example of some of the styles referenced above:

Screen Shot 2019-10-11 at 2 59 24 PM

It looks like sometimes the button gets grouped with the button next to it inside a `div` because the DOM structure of the buttons can differ depending on the location. Again, would advocate for landing this and refining that further in the future!

ext/src/content/content.js
```diff
@@ -52,89 +44,131 @@ function init() {
...
-        style
-      });
-      activeButtons.set(target, buttonEl);
-      buttonEl.addEventListener("click", event => {
-        event.preventDefault();
-        target.value = window.prettier.format(target.value, {
-          parser: "markdown",
-          plugins: window.prettierPlugins
+  function discoverButtonsAndCreatePrettierButtons() {
+    const BUTTON_STYLE = { float: "left", "margin-right": "10px" };
```
kaicataldo: I think where the button is rendering and using `float` here is causing some of the wonkiness with styles. It would be great if we could render it in the `div` containing the other buttons (the original iteration did this well, I think). Again, not a blocker, but leaving this comment for posterity and for the person who will work on this in the future.

Wicked7000: Agreed, the issue is mainly that a lot of different buttons have different divs they are contained in (eg. 'Close pull request' and 'Comment' buttons are actually within separate div elements) but can look into styling that gives consistent results across different sets of buttons.

kaicataldo:
Would be great to land this! This is a huge improvement over the current version!

kaicataldo:
@lipis Any chance we can merge this? I think we can probably do something similar to fix https://github.com/prettier/prettier-chrome-extension/issues/12.

Unrelated, but I'm happy to help maintain this project. This is something I've been wanting to implement anyway and I have some bandwidth to both contribute and help move PRs along :)

lipis:
> Unrelated, but I'm happy to help maintain this project. This is something I've been wanting to implement anyway and I have some bandwidth to both contribute and help move PRs along :)

You are more than welcome! I sent you an invite.. :)

kaicataldo:
Thanks!

PR #23, opened 2019-10-13
Bundle with webpack

### LINKED ISSUE #8: Figure out how to include Prettier files on release
It probably makes sense to just use npm to install Prettier and bundle the injected content script using something like Webpack.

nickmccurdy:
Fixes #8 

- [x] Bundle dependencies
- [x] Install Prettier via npm instead of vendoring
- [x] Optimize extension size
  - [x] Use webpack's production mode (Chrome currently errors saying the bundle isn't UTF8 encoded)


nickmccurdy:
The bundle size is still an alarmingly large 8MB and Prettier doesn't seem to have an ES module build for browsers with code splitting support. We might need to experiment with loading parsers dynamically or use a CDN rather than webpack. Any opinions of offline support VS using a CDN to make it smaller?

nickmccurdy:
> One possible interesting feature that loading dynamically could allow would be matching the Prettier version and reading Prettier configuration from a respective GitHub repository’s code, but there’s obviously a lot of basic functionality we should implement before we can get to that.

Unfortunately the [browser build of Prettier](https://prettier.io/docs/en/browser.html) doesn't support configuration files, and I don't think bundling the Node build would work as it relies on native libraries like `fs` which don't exist in Chrome extensions. However it seems like the config file format is a subset of the options allowed in the `prettier.format` API, so we could parse them manually and merge then into the parser/plugin options using the browser build.

nickmccurdy:
It seems like Chrome recommends against using CDNs to load scripts dynamically for security (https://developer.chrome.com/extensions/contentSecurityPolicy#resourceLoading), so webpack is probably still the most convenient way to manage our dependencies. I also don't see a point in disabling some parsers if we're going to bundle all of them anyway. I'll try to find other ways to reduce the bundle size, but it's difficult with Prettier's lack of ES bundles for browsers, and I've seen extensions much larger than this.

kaicataldo:
That makes sense to me! This definitely seems like a better way of managing Prettier than the current way (and what I was expecting we would eventually do when I copied in those files initially). I think this is a really great start, and we can continue to optimize things as we go.

webpack.config.js
```diff
@@ -0,0 +1,9 @@
+const path = require("path");
+
+module.exports = {
+  entry: "./ext/src/content/content",
+  mode: "development",
```
kaicataldo: Would it allow for more flexibility if we don’t set the `mode` here? Will `npm run build` be running in `development` mode?

package.json
```diff
@@ -25,15 +26,19 @@
     "prettier": "prettier \"**/*.{md,css,html,json,yml}\" --write",
     "test:other": "yarn prettier --list-different",
     "test:js": "eslint --ignore-path .gitignore --ignore-path .prettierignore \"**/*.js\"",
-    "test": "yarn test:other && yarn test:js"
+    "test": "yarn test:other && yarn test:js",
+    "build": "webpack"
```
kaicataldo: Could it make sense to remove setting mode to `development` in the webpack config and instead add a `dev` script that runs `webpack --mode=development` (leaving the `build` script as is since I believe it runs in `production` mode by default).

kaicataldo:
> Use webpack's production mode (Chrome currently errors saying the bundle isn't UTF8 encoded)

Sorry. Missed this. That’s very strange...I think it would be fine to merge this as-is and update that later, as long as we make a follow-up issue so we don’t lose track of it.

webpack.config.js
nickmccurdy: If you don't set `mode` at all, it defaults to `production` which causes Chrome to error when installing the extension because of a non-UTF8 file encoding. I've tried several webpack plugins that force the encoding on the build and they don't seem to work. I'm not sure how to fix it, but we might be able to get away with just some of the performance optimizations by using `mode: "none",` and manually enabling settings that don't cause issues with Chrome.

nickmccurdy:
> Just a small idea about adding a `dev` npm script in addition to `build`, but otherwise this LGTM. Thanks!

I use `yarn build --watch` for this, but I can add `yarn dev` as an alias if you prefer.

package.json
nickmccurdy: This would work, but I'd prefer to keep all the configuration in the webpack config unless we need to have multiple build scripts for different modes (currently `production` doesn't work anyway and the only advantage of `none` is speed).

nickmccurdy:
> > Use webpack's production mode (Chrome currently errors saying the bundle isn't UTF8 encoded)
> 
> Sorry. Missed this. That’s very strange...I think it would be fine to merge this as-is and update that later, as long as we make a follow-up issue so we don’t lose track of it.

Sure, I'd be comfortable merging this without further optimizations (and creating a separate follow up issue) if you are.

kaicataldo:
I'm okay with it - let's see what @lipis says.

nickmccurdy:
I've managed to fix the production mode build by disabling webpack's minification. This increases the production bundle size from 5.2 megabytes to 6.5 megabytes, but fixes Chrome extension installation while still shaving 1.5 megabytes off the development bundle.

kaicataldo:
Mind fixing up the merge conflicts?

nickmccurdy:
Forgot about that, done

kaicataldo:
Very sorry, but mind fixing up the merge conflicts once more? I ended up touching some of the same lines in my last PR. Might be easier to just checkout the the content/content.js file in master and reapply the changes :)

nickmccurdy:
It's okay, I fixed it.

kaicataldo:
Thanks for contributing!

fisker:
got attention from https://github.com/prettier/prettier/issues/6656

just curious why not rollup？

fisker:
IMO, rollup is more common for js libs, treeshaking? maybe

Need compare the build result to tell

nickmccurdy:
Personally I only find it necessary to use Rollup when we're emitting ES modules, and I don't think Chrome extensions support that.

Issue #22, opened 2019-10-13
Formatting on GitHub interrupts undo history


nickmccurdy:
When I type some text in a supported field on GitHub, clicking the Prettier formatting button prevents me from undoing some or all of the changes to the text field, starting from the most recent. This may confuse users that decide to undo something before formatting, and also makes it difficult to recover from unwanted formatting changes.

nickmccurdy:
Fixed by #22

PR #24, opened 2019-10-13
Maintain Browser history on 'Prettier' + Fix issue with button on appearing on 'Create pull Request'

Wicked7000:
This PR is to address the issue raised in:
#22 

This was fixed by switching to document.execCommand to insert the text into the textarea, which allows the browser to continue to keep track of history -> We should probably use this as the standard way of modifying text rather than setting the value.

Along with a small bugfix to check for the existence of the 'js-discussion' element before trying to observe it. (This was causing the 'Prettier' button to not render on the Create pull request pages)

Let me know if you have any questions or concerns.

### DIFF
--- a/ext/src/content/content.js
+++ b/ext/src/content/content.js
@@ -116,11 +116,14 @@ function init() {
         const textArea = findWithClass(buttonElem, "comment-form-textarea");
         buttonElem.addEventListener("click", event => {
           event.preventDefault();
-          textArea.value = window.prettier.format(textArea.value, {
+          const formattedText = window.prettier.format(textArea.value, {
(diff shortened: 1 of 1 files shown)


nickmccurdy:
For future reference if you add `fix #ISSUE_ID` or `fixes #ISSUE_ID` to the pull request description or a commit message, GitHub will automatically close the issue when the pull request is merged into master.

Wicked7000:
Ah will do so in future, Thanks! I'm relatively new to open source so apologies for not knowing these things

lipis:
No worries @Wicked7000 :) We've all been there.. welcome.

PR #26, opened 2019-10-14
Fix stack overflow buttons

### LINKED ISSUE #12: Fix button rendering behavior for logged out commenting on Stack Overflow

### LINKED ISSUE #18: Show prettier button when editing questions or answers on StackOverflow
Right now the button appears only when writing new questions or answers.

Wicked7000:
Fixes #18 
Fixes #12 

The following PR makes some minor refactoring changes to the github code to make it slightly cleaner. It also makes it so the Prettier button appears on all sections on StackOverflow, along with fixing the logged out issue.

Caveats: 
- Maintaing undo history is not very simple for the stack overflow editor, stackoverflow uses a wmd editor which they have made a fork [pagedown](https://github.com/StackExchange/pagedown/blob/master/Markdown.Editor.js). This editor handles state based on keypress and statechange and due to some extension limitations sending a keydown event with custom properties is not very easy. For now I have used the paste event which works sort of but can be buggy based on when the event actually reaches their editor code. This might have to be something we look into in the future to see if there is something we could use to fix it.


ext/src/content/content.js
```diff
@@ -340,42 +211,200 @@ function init() {
...
+          formattedBlock = isCodeBlock
+            ? `${firstLine}\n${formattedBlock}\n${lines[lines.length - 1]}`
+            : `${firstLine}\n${lines[1]}\n    ${
+                formattedBlock.split("\n").length > 1
+                  ? formattedBlock.split("\n").join("\n    ")
+                  : formattedBlock
+              }`;
+
+          const value = inputEl.value.replace(lines.join("\n"), formattedBlock);
+          stackOverflowFormat(event, inputEl, value, null);
```
kaicataldo: It looks like code is being formatted twice now: [here](https://github.com/prettier/prettier-chrome-extension/pull/26/files#diff-276852ee1731302f182d2bc97d60c46fR357) and [here](https://github.com/prettier/prettier-chrome-extension/pull/26/files#diff-276852ee1731302f182d2bc97d60c46fR34).

I wonder if it would make sense to actually build up the formatted string (saving off the initial value to a variable in memory and then replacing the values during this `Array.prototype.forEach()` iteration) and after running all the transformations of the text just pasting it in once? It would also have the side effect of not having multiple paste events. Currently, Ctrl + Z will only go back one paste, so if you have multiple code blocks you have to Undo multiple times to get back to the pre-formatted state. Thoughts?

Seems like this whole callback starting on line 250 should be the "stackOverflowFormat()` function.

Wicked7000: Yeah that makes sense I can make that change :) 

Although the problems with the stack overflow editor still lead to slightly buggy behaviour. (As paste events aren't always causing the editor to save state). I thinks it's best to leave that for another issue as it isn't very straightforward and prior to this the stack overflow integration wasn't working.

lipis:
is it ok to merge?

kaicataldo:
I'm fine with merging it and fixing it up in a followup PR. Looks like we just need to fix up the merge conflict.

lipis:
Let's wait for @Wicked7000 then..

Wicked7000:
I've actually re-written most of the code block parsing for stack overflow as it wasn't very clear and it didn't seem to be working. (i.e it didn't work in cases when text was outside any block) this should account for that better now without using the reduce function. (Which I personally find rather confusing when reading code)

Let me know your opinions or if there are any other bugs with it, I've tested quite a bit and it seems to work well. (I've also done some more refactoring based on previous comments)

lipis:
🍻

ext/src/content/content.js
```diff
@@ -417,42 +221,242 @@ function init() {
...
+        if (block.codeBlock) {
+          arrToAdd.unshift(block.firstLine);
+          if (block.lastLine) {
+            arrToAdd.push(block.lastLine);
+          }
+        }
+        formattedText.push(...arrToAdd);
+      });
+    } else {
+      markdownText = prettier.format(inputEl.value, {
```
kaicataldo: The previous iteration always ran Prettier using the markdown editor on the entire value after running Prettier on the individual code blocks. I think it would still be nice to do this, because it formats outside the code blocks as well (since the markdown parser doesn't format the StackOverflow-specific code block syntax).

ext/src/content/content.js
```diff
@@ -417,42 +221,242 @@ function init() {
...
+  // TODO: This solution is buggy and does not always work - more discussion needed on solution
+  function stackOverflowFormat(event, inputEl) {
+    event.preventDefault();
+    const codeBlocks = getCodeBlocksStackOverflow(inputEl);
+
+    let markdownText = inputEl.value;
+    const formattedText = [];
+    if (codeBlocks.length) {
+      /*
+       * TODO: Add support for language-all: <!-- language-all: lang-* -->
```
kaicataldo: Any idea where [this](https://github.com/prettier/prettier-chrome-extension/commit/0407f000ce6632eb2fc54adb7b2672d8efafc0b6) went? Looks like a bad merge conflict resolution.

kaicataldo:
Looks like there was a bad merge conflict - some of the behavior merged in today has disappeared.

lipis:
Oups.. I merged it too fast!.. Had to wait a bit.. feel free to revert or re add what's missing.

lipis:
Off to bed.. it's a bit late 😴 🛌

kaicataldo:
> I've actually re-written most of the code block parsing for stack overflow as it wasn't very clear and it didn't seem to be working. (i.e it didn't work in cases when text was outside any block)

Can you expand on this a bit? I'm not sure I understand. I tested it quite a bit and it was working fine. If you're referring to code snippets (not blocks), support for that was added [here](https://github.com/prettier/prettier-chrome-extension/commit/0407f000ce6632eb2fc54adb7b2672d8efafc0b6). It wasn't in the original implementation because we didn't support `language-all` comments and therefore couldn't know what language to use.

ext/src/content/content.js
```diff
@@ -417,42 +221,242 @@ function init() {
...
+      if (emptyStringRegex.test(line)) {
+        continue;
+      }
+
+      const langIndented = getIndentedCodeBlock(line);
+      const codeBlock = getCodeBlock(line);
+      const lang = codeBlock || langIndented;
+
+      if (lang) {
+        if (outsideOfBlocks.length) {
```
kaicataldo: Do you mind clarifying what "outside of blocks" refers to?

kaicataldo:
Formatting indented code blocks is broken for me now, and the formatter is removing blank lines in places it wasn't before. I think we should revert this for now.

kaicataldo:
refs https://github.com/prettier/prettier-chrome-extension/pull/32

I think it would be best to split this into two PRs. Totally fine to refactor the block finding logic (might be nice to actually abstract that out into its own module and then write some tests since it should be quite easy to test transforming the string input), but it feels like this PR is doing a bit too much at the moment. @Wicked7000 How does that sound to do?

kaicataldo:
Ah, I see what you mean about it being broken. I must have accidentally removed the change to initialize this variable correctly :/ See https://github.com/prettier/prettier-chrome-extension/pull/34

kaicataldo:
@Wicked7000 Please feel free to reopen this, and apologies for the confusion here.

Wicked7000:
I'll separate this into two separate PRs and explain my refactoring a little better and see if you guys agree with it, If not I'm more than happy to scrap it and continue using what's already there if it's working well!

kaicataldo:
Thanks, and appreciate your flexibility! Definitely not tied to the current implementation, but don't want to lose any progress we've re features. Splitting this out into two separate PRs would be great!

Issue #13, opened 2019-10-09
Add support for language-all comments in Stack Overflow


kaicataldo:
Stack Overflow allows for `language-all` comments (see [here](https://stackoverflow.com/editing-help#syntax-highlighting) for more details) in the format of `<!-- language-all: lang-html -->`. We should add support for this!

nickmccurdy:
Do you think it would be enough to convert to GFM, format, and then convert back? Or does the formatting need to happen with a data structure specifically designed around Stack Overflow's parsing?

kaicataldo:
Given the current implementation, and the way it goes through processing code blocks in the order in which they appear, I think it would be easier to track the language declared in the last `language-all` comment and apply it when Prettier runs. For some context, the current code currently uses regexes to find these code blocks and replaces them in the code, which seemed like a simpler solution to me than trying to convert formats and convert back. 

That being said, the current implementation is just the first iteration I wrote to get the basic functionality working, and there's lots of room for improvement! I'm sure it'll change a lot as time goes on :)

PR #27, opened 2019-10-15
Feature: respect language-all comments for StackOverflow

kaicataldo:
fixes https://github.com/prettier/prettier-chrome-extension/issues/13

Please let me know what you think of this. It currently uses the last language defined by ` language-all` comment if a comment doesn't have a language attribute.

### DIFF
--- a/ext/src/content/content.js
+++ b/ext/src/content/content.js
@@ -3,6 +3,7 @@
 function init() {
   const GITHUB_URL = "https://github.com";
   const GITHUB_VALID_PATHNAMES = /^\/.*\/.*\/(?:pull\/\d+(?:\/?|\/files\/?)$|commit|compare\/.*|issues\/\d+|issues\/new)/u;
+  const POLLING_INTERVAL = 30;
   let isGithubListenerAdded = false;
 
   const STACKOVERFLOW_URL = "https://stackoverflow.com";
@@ -190,6 +191,7 @@ function init() {
         const codeBlocks = inputEl.value.split("\n").reduce((groups, line) => {
           const codeBlockRegex = /^\s{0,3}(?:```|~~~)/u;
           const indentedCodeLangRegex = /^\s*<!-- language: lang-.* -->/u;
+          const langAllRegex = /^\s*<!-- language-all: lang-.+ -->/u;
           const emptyLineRegex = /^\s*$/u;
           const indentedCodeBlockRegex = /^\s{4}/u;
           const codeSnippetRegex = /`{1,2}[^\n`]+`{1,2}/u;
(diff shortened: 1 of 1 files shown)


ext/src/content/content.js
```diff
@@ -3,6 +3,7 @@
 function init() {
   const GITHUB_URL = "https://github.com";
   const GITHUB_VALID_PATHNAMES = /^\/.*\/.*\/(?:pull\/\d+(?:\/?|\/files\/?)$|commit|compare\/.*|issues\/\d+|issues\/new)/u;
+  const POLLING_INTERVAL = 30;
```
kaicataldo: Added this back because it's referenced later in the code. Not sure why CI didn't catch this.

ext/src/content/content.js
```diff
@@ -268,50 +277,95 @@ function init() {
...
+                codeSnippetRegex,
+                (match, openingBackticks, snippet, closingBackticks) => {
+                  let formattedSnippet = snippet;
+
+                  try {
+                    formattedSnippet = window.prettier.format(snippet, {
+                      parser: PARSERS_LANG_MAP[lang],
+                      plugins: window.prettierPlugins
+                    });
+                  } catch {}
```
kaicataldo: Wrapping this in a `try`/`catch` statement allows the plugin to continue onto the next code snippet or block without crashing. This will probably happen frequently since Prettier throws if the parser can't parse something.

ext/src/content/content.js
```diff
@@ -268,50 +277,95 @@ function init() {
...
+            let formattedText = lines.join("\n");
+
+            // Code Snippets
+            if (codeSnippetRegex.test(firstLine)) {
+              formattedText = firstLine.replace(
+                codeSnippetRegex,
+                (match, openingBackticks, snippet, closingBackticks) => {
+                  let formattedSnippet = snippet;
+
+                  try {
```
kaicataldo: Wrapping this in a `try`/`catch` statement allows the plugin to continue onto the next code snippet or block without crashing. This will probably happen frequently since Prettier throws if the parser can't parse something.

ext/src/content/content.js
```diff
@@ -268,50 +277,95 @@ function init() {
...
+              if (isCodeBlock) {
+                codeLines = codeLines.slice(1, -1);
+              } else {
+                const indentedLineCodeBlockStartIdx = 2;
+                codeLines = isIndentedBlockWithLang
+                  ? lines.slice(indentedLineCodeBlockStartIdx)
+                  : lines;
+              }
+
+              try {
```
kaicataldo: Wrapping this in a `try`/`catch` statement allows the plugin to continue onto the next code snippet or block without crashing. This will probably happen frequently since Prettier throws if the parser can't parse something.

PR #35, opened 2019-10-16
Chore: clean up file structure and bundle setup

kaicataldo:
This gets rid of currently unused files and change the directory structure a bit to (I hope) make things clearer!


nickmccurdy:
I haven’t had a chance to test this yet, but based on the code, it seems like this will use production mode with minification which makes the extension uninstallable in Chrome.

kaicataldo:
I remember you mentioned that, but I just tested it locally and it worked fine for me, so not sure what's going on there. Any ideas what might be causing that? I'm sure if you're running it others will as well.

kaicataldo:
Scratch that. I just tested it again and I'm getting the same thing. Wonder why it didn't show up before...

nickmccurdy:
Note that `webpack —watch` does not update the config file if it is changed

kaicataldo:
I don't think that was the issue - might have just been a stale artifact. Will either revert that change or see if I can fix it some other way. Thanks!

kaicataldo:
I believe I've fixed it - PR here! https://github.com/prettier/prettier-chrome-extension/pull/36

package.json
```diff
@@ -4,31 +4,17 @@
...
     "fix:other": "yarn prettier --write",
     "fix:js": "yarn test:js --fix",
     "fix": "yarn fix:js && yarn fix:other",
     "install": "yarn build",
     "prettier": "prettier \"**/*.{md,css,html,json,yml}\" --write",
     "test:other": "yarn prettier --list-different",
     "test:js": "eslint --ignore-path .gitignore --ignore-path .prettierignore \"**/*.js\"",
-    "test": "yarn test:other && yarn test:js",
-    "build": "webpack"
+    "test": "yarn test:other && yarn test:js"
```
lipis: maybe sorting by name is the best here..

kaicataldo: Sure, I don't have a strong opinion either way

PR #36, opened 2019-10-16
Chore: minify bundle

kaicataldo:
`terser-webpack-plugin` is the default minification plugin and is already included as a dependency of webpack. Generally, I advocate for explicitly listing dependencies, but in this case, I think it makes sense to make sure we're using the version installed by webpack.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,3 @@
-extension.zip
-extension/dist
 node_modules
+extension/dist
+extension.zip
--- a/extension/src/content/index.js
+++ b/extension/src/content/index.js
@@ -1,8 +1,5 @@
-import parserAngular from "prettier/parser-angular";
 import parserBabylon from "prettier/parser-babylon";
 import parserFlow from "prettier/parser-flow";
-import parserGlimmer from "prettier/parser-glimmer";
-import parserGraphql from "prettier/parser-graphql";
 import parserHtml from "prettier/parser-html";
 import parserMarkdown from "prettier/parser-markdown";
 import parserPostcss from "prettier/parser-postcss";
@@ -12,11 +9,8 @@ import prettier from "prettier/standalone";
 
 function init() {
   const prettierPlugins = [
-    parserAngular,
     parserBabylon,
     parserFlow,
-    parserGlimmer,
-    parserGraphql,
     parserHtml,
     parserMarkdown,
     parserPostcss,
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,7 +1,22 @@
 const path = require("path");
+const TerserPlugin = require("terser-webpack-plugin"); // included as a dependency of webpack
 
 module.exports = {
   entry: "./extension/src/content/index.js",
+  mode: "production",
+  optimization: {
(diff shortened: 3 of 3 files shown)


kaicataldo:
@nickmccurdy Mind taking this for a spin and letting me know if it works for you?

webpack.config.js
```diff
@@ -1,7 +1,19 @@
...
 
 module.exports = {
   entry: "./extension/src/content/index.js",
+  optimization: {
+    minimize: true,
+    minimizer: [
+      new TerserPlugin({
+        terserOptions: {
+          // https://github.com/webpack-contrib/terser-webpack-plugin/issues/107
+          output: { ascii_only: true }
```
nickmccurdy: I tried this myself before but webpack kept generating an extra file for some reason

.gitignore
```diff
@@ -1,3 +1,4 @@
 extension.zip
 extension/dist
 node_modules
+package-lock.json
```
nickmccurdy: npm shouldn't be used at all since this project uses yarn's version resolution, I think it's better to explicitly show the npm lockfile as untracked to remind contributors to use yarn

webpack.config.js
kaicataldo: Fixed!

.gitignore
kaicataldo: That's fair. Will remove.

extension/src/content/index.js
```diff
@@ -1,8 +1,5 @@
-import parserAngular from "prettier/parser-angular";
```
kaicataldo: Removing currently unused parsers -- saves us a couple MB in the bundle :)

kaicataldo:
Thanks for reviewing!

extension/src/content/index.js
nickmccurdy: Isn't it possible for other parsers to be used dynamically?

kaicataldo: Most likely? Since these parsers aren't in the map that's being checked, this is currently dead code. Happy to add them back if we start supporting those languages. The glimmer (Handlebars) parser seems to need another kind of loader though.

nickmccurdy: Oh now I understand, PARSERS_LANG_MAP selects the parser so other parsers unsupported by Stack Overflow (except markdown) wouldn't be used. Don't we still need glimmer/handlerbars and graphql support for GitHub though?

kaicataldo: Ack, you're right. I was a little too quick to remove those. I'll have to test if Glimmer actually works or not because we're getting webpack errors when trying to load that parser module.

nickmccurdy: That should be fixed by Prettier 1.19, see https://github.com/prettier/prettier/issues/6656.

I don't really use Glimmer/Handlebars/Ember, but if you can verify that they're broken I'd be fine with removing the parser until then.

kaicataldo: Since we’re still essentially doing pre-releases, I think it’s fine as-is. We should ensure we update it before we release v1 though!

PR #37, opened 2019-10-16
Fix: fix css parser warning in console

kaicataldo:
Screen Shot 2019-10-16 at 1 10 14 AM

This PR gets rid of this warning and adds a new mapping for `lang-sass` (in addition to `lang-scss` for Stack Overflow.


nickmccurdy:
Does this cause the same warning when using the postcss processor with less/sass/scss?

kaicataldo:
I'm not seeing the same warning for those.

PR #40, opened 2019-10-16
Chore: lock all package versions

kaicataldo:
~~Realized that we had locked the versions of a lot of our deps. This PR updates the semver range so to allow for semver-minor upgrades and updates two deps that have had a new semver-patch version released.~~

~~The issue I found of webpack warning about missing `require.extensions` should be fixed in the next Prettier release! (see https://github.com/prettier/prettier/issues/6656 for more details).~~

### DIFF
--- a/package.json
+++ b/package.json
@@ -25,8 +25,8 @@
     "eslint-plugin-prettier": "3.1.1",
     "husky": "3.0.8",
     "lint-staged": "9.4.2",
-    "webpack": "^4.41.1",
-    "webpack-cli": "^3.3.9"
+    "webpack": "4.41.1",
+    "webpack-cli": "3.3.9"
   },
   "lint-staged": {
     "*.{js}": [
--- a/yarn.lock
+++ b/yarn.lock
@@ -4036,7 +4036,7 @@ watchpack@^1.6.0:
     graceful-fs "^4.1.2"
     neo-async "^2.5.0"
 
-webpack-cli@^3.3.9:
+webpack-cli@3.3.9:
   version "3.3.9"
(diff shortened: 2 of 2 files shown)


kaicataldo:
Doesn't the lockfile fix this, though? Any package changes are explicit.

kaicataldo:
To be clear, I ran `yarn upgrade husky` and `yarn upgrade webpack` after updating the semver ranges.

lipis:
There is no real gain in having the ranges. Let's pin the versions and we have the dependabot to keep us updated..

kaicataldo:
I didn't know we had a bot that could let us know about dependency upgrades - interesting!

nickmccurdy:
I find it counterproductive to lock versions in package.json. They're already locked by the yarn.lock file, but also locking them in package.json means that useful tools like dependabot and `yarn upgrade-interactive` can't properly upgrade semver-compatible releases.

We should add back `^` to each version. As long as you're using yarn, versions will still be locked and consistently on all systems.

kaicataldo:
I personally usually use semver ranges to better communicate what I think is acceptable re dependencies (and, as you say, tooling can then behave in line with my intentions) and rely on lock files to ensure consistent package versions. That being said, I’m happy to defer to what everyone else thinks here, since I don’t think it matters as much in the context of a Chrome Extensions that probably won’t need a ton of maintenance once it is feature complete.

nickmccurdy:
Also one thing I forgot to mention, `yarn add` and `npm install` default to using `^` for new package installations. So if you're already using that format, new installs will be consistent.

PR #41, opened 2019-10-16
Add parsers

kaicataldo:
This PR adds back the removed parsers that @nickmccurdy correctly [pointed out](https://github.com/prettier/prettier-chrome-extension/pull/36#discussion_r335286791) are still needed by the GitHub formatter.

### DIFF
--- a/extension/src/content/index.js
+++ b/extension/src/content/index.js
@@ -1,5 +1,8 @@
+import parserAngular from "prettier/parser-angular";
 import parserBabylon from "prettier/parser-babylon";
 import parserFlow from "prettier/parser-flow";
+import parserGlimmer from "prettier/parser-glimmer";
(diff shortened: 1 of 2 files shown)


nickmccurdy:
Out of curiosity, do you know what the Angular parser is for? Do they have their own file format like Vue now? I haven't used it since AngularJS 1, I thought it still used ordinary JS files and HTML files with special attributes.

kaicataldo:
Taking a cursory look, it looks like it’s used for the HTML parser (see [here](https://github.com/prettier/prettier/blob/ff79721397aa8379439c216959a455cee5b5af8c/src/language-html/parser-html.js)).

nickmccurdy:
In that case I guess it's useful to keep it, I just wanted to make sure it's being used.

PR #45, opened 2019-10-16
Use recommended version ranges

nickmccurdy:
Fixes issues described in https://github.com/prettier/prettier-chrome-extension/pull/40#issuecomment-542898672.

Note that the actual versions currently used in the `yarn.lock` are exactly the same, this just improves how they can be upgraded in the future (which requires manual intervention or some external tool like dependabot).

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,16 +17,16 @@
     "test:js": "eslint --ignore-path .gitignore --ignore-path .prettierignore \"**/*.js\""
   },
   "dependencies": {
-    "prettier": "1.18.2"
+    "prettier": "^1.18.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


kaicataldo:
This is how I personally prefer to manage dependencies, as I think it expresses intention better. That being said, I’m happy to defer to consensus.

I made a PR that did this same thing earlier and we ended up deciding not to merge it. I know you’ve already seen the discussion, but leaving it here for posterity: https://github.com/prettier/prettier-chrome-extension/pull/40#issuecomment-542905615

nickmccurdy:
Oh, it was originally adding the ranges? I didn't notice.

PR #49, opened 2019-10-17
Fix issues related to StackOverflow buttons

### LINKED ISSUE #12: Fix button rendering behavior for logged out commenting on Stack Overflow

### LINKED ISSUE #18: Show prettier button when editing questions or answers on StackOverflow

Wicked7000:
(Re-creation of a previous PR with separated changes, let me know any further comments)

- On the comment about stackOverflowFormat containing the entire listener I think I'll leave that for the other PR where I do more formatting anyways as the listener in general is too big to be in one function.

Fixes #18
Fixes #12

The following PR makes some minor refactoring changes to the github code to make it slightly cleaner. It also makes it so the Prettier button appears on all sections on StackOverflow, along with fixing the logged out issue.

Also adds main.js to .gitignore -> as I assume we don't want people committing that

Caveats:
Maintaing undo history is not very simple for the stack overflow editor, stackoverflow uses a wmd editor which they have made a fork pagedown. This editor handles state based on keypress and statechange and due to some extension limitations sending a keydown event with custom properties is not very easy. For now I have used the paste event which works sort of but can be buggy based on when the event actually reaches their editor code. This might have to be something we look into in the future to see if there is something we could use to fix it.

### DIFF
--- a/extension/src/content/index.js
+++ b/extension/src/content/index.js
@@ -26,11 +26,10 @@ function init() {
 
   const GITHUB_URL = "https://github.com";
   const GITHUB_VALID_PATHNAMES = /^\/.*\/.*\/(?:pull\/\d+(?:\/?|\/files\/?)$|commit|compare\/.*|issues\/\d+|issues\/new)/u;
-  const POLLING_INTERVAL = 30;
   let isGithubListenerAdded = false;
 
   const STACKOVERFLOW_URL = "https://stackoverflow.com";
-  const STACKOVERFLOW_VALID_PATHNAMES = /^\/questions/u;
+  const STACKOVERFLOW_VALID_PATHNAMES = /(^\/questions|\/posts\/\d+\/edit)/u;
   const PARSERS_LANG_MAP = {
     css: "css",
     flow: "flow",
@@ -70,29 +69,7 @@ function init() {
     return button;
   }
 
-  function setupCommentObserver(observer) {
-    const JS_DIFF_CONTAINER_CLASS = ".js-diff-progressive-container";
-    const COMMENT_TIMELINE_GROUP_CLASS = ".timeline-comment-group";
-
-    for (const elem of document.querySelectorAll(
-      COMMENT_TIMELINE_GROUP_CLASS
-    )) {
-      observer.observe(elem, {
-        attributes: true,
-        childList: true,
-        subtree: true
-      });
-    }
-
-    for (const elem of document.querySelectorAll(JS_DIFF_CONTAINER_CLASS)) {
-      observer.observe(elem, {
-        childList: true,
-        subtree: true
-      });
-    }
-  }
-
-  function searchAndAddListenerToButtons() {
+  function seachForGithubButtons() {
     const COMMENT = "Comment";
     const REPLY = "Reply…";
     const CANCEL = "Cancel";
@@ -128,7 +105,7 @@ function init() {
       }
       if (button.innerText === REPLY) {
         const observer = new MutationObserver(() => {
-          discoverButtonsAndCreatePrettierButtons();
+          createGithubPrettierButtons();
         });
         observer.observe(
           findWithClass(button, "inline-comment-form-container"),
@@ -139,9 +116,9 @@ function init() {
     return createList;
   }
 
-  function discoverButtonsAndCreatePrettierButtons() {
+  function createGithubPrettierButtons() {
     const BUTTON_STYLE = { float: "left", "margin-right": "10px" };
-    const createList = searchAndAddListenerToButtons();
+    const createList = seachForGithubButtons();
 
     for (const button of createList) {
       let prettierBtn = button.parentNode.querySelector(".prettier-btn");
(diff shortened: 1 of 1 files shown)


extension/src/content/index.js
```diff
@@ -186,228 +203,28 @@ function init() {
...
-        }
-
-        if (codeBlocks.length) {
-          let langAll = null;
-
-          // https://stackoverflow.com/editing-help#syntax-highlighting
-          codeBlocks.forEach(lines => {
-            const codeBlockRegex = /^\s{0,3}(?:```|~~~)\s*(?:lang-(.+))?/u;
-            const indentedCodeWithLangRegex = /^\s*<!-- language: lang-(.+) -->/u;
-            const langAllRegex = /^\s*<!-- language-all: lang-(.+) -->/u;
```
kaicataldo: Looks like we're missing the `language-all` and code snippet formatting behavior that was added [here](https://github.com/prettier/prettier-chrome-extension/commit/0407f000ce6632eb2fc54adb7b2672d8efafc0b6). Maybe from a bad merge conflict resolution? Might be worth copying over the entire click event handler from `master` starting from https://github.com/prettier/prettier-chrome-extension/pull/49/files#diff-2b5769826b6670c8ab28f63985a301a8L210.

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 extension/dist
 extension.zip
+ext/main.js
```
kaicataldo: The current version in `master` outputs the bundle in `extension/dist`, so this should be fine as is! Maybe double check your branch is rebased off of latest `master`?

extension/src/content/index.js
```diff
@@ -46,6 +45,42 @@ function init() {
...
+    text = null,
+    parser = "markdown"
+  ) {
+    event.preventDefault();
+    textAreaElem.dispatchEvent(new Event("paste"));
+    textAreaElem.input = formatText(event, textAreaElem, text, parser);
+    textAreaElem.dispatchEvent(new Event("paste"));
+  }
+
+  function formatText(event, textAreaElem, text = null, parser = "markdown") {
```
kaicataldo: This function has a really complex/confusing signature for what it's doing, which I think is a code smell indicating it might be doing too much. I actually think it's simpler to have the GitHub and StackOverflow formatting functions be separated out for now (my plan is to move them into separate modules once we land this) rather than coupled like this. We can always refactor and create abstractions later as we find similar code. Thoughts?

extension/src/content/index.js
```diff
@@ -46,6 +45,42 @@ function init() {
     yaml: "yaml"
   };
 
+  // Pasting events are used to force stackoverflows WMD to save state.
+  // TODO: This solution is buggy and does not always work - more discussion needed on solution
+  function stackOverflowFormat(
```
kaicataldo: This is only doing some formatting - the the bulk of the formatting is happening [here](https://github.com/prettier/prettier-chrome-extension/pull/49/files#diff-2b5769826b6670c8ab28f63985a301a8R276-R401).

Can we not abstract this out yet? Or have it be just the DOM mutation?

extension/src/content/index.js
kaicataldo: I do think it makes sense to potentially share code for the actual DOM mutation though, since there is some shared behavior around focusing the text area, etc.

Wicked7000:
I'll rebase this and modify the format function later today sorry for the wait

lipis:
what's wrong with Travis...?

Wicked7000:
Mentioned this on the other PR but it's possible it's this?

Possibly could be this?
https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154

extension/src/content/index.js
```diff
@@ -26,11 +26,10 @@ function init() {
 
   const GITHUB_URL = "https://github.com";
   const GITHUB_VALID_PATHNAMES = /^\/.*\/.*\/(?:pull\/\d+(?:\/?|\/files\/?)$|commit|compare\/.*|issues\/\d+|issues\/new)/u;
-  const POLLING_INTERVAL = 30;
   let isGithubListenerAdded = false;
 
   const STACKOVERFLOW_URL = "https://stackoverflow.com";
```
Wicked7000: this is a _tesT_

Wicked7000: this _is_ a test

lipis:
Can we fix the conflicts and get that in?

kaicataldo:
Would be great to land this! Currently holding off on further work to avoid further conflicts.

extension/src/content/index.js
```diff
@@ -26,11 +26,10 @@ function init() {
...
-  const STACKOVERFLOW_VALID_PATHNAMES = /^\/questions/u;
+  const STACKOVERFLOW_VALID_PATHNAMES = /(^\/questions|\/posts\/\d+\/edit)/u;
```
Wicked7000: t

Wicked7000:
Okay so I've re-based and everything seems to be working fine now, with both Github prettier and StackOverflow buttons.

A few points:
- Stackoverflow comments don't work yet (I thought this could be seperated out into its own PR / Issue for now)
- The undo history for stackoverflow (pasting code) has been left out for now as it was reliable anyways that should probably be split into its own PR / Issue too.

Let me know if there is any bugs you experience, I've tried to test for most cases I've managed to find.

nickmccurdy:
Seems like Travis is working again

kaicataldo:
Thanks!

PR #50, opened 2019-10-17
Add type=`button` to rendered prettier buttons

### LINKED ISSUE #44: Button is missing `type="button"` attribute
This means it may submit the form it lives inside, which is probably not desired…

Screen Recording 2019-10-16 at 18 15 12

Wicked7000:
Fixes #44 
Simple change to add the type='button' to our rendered Prettier buttons.

### DIFF
--- a/extension/src/content/index.js
+++ b/extension/src/content/index.js
@@ -52,6 +52,7 @@ function init() {
   ) {
     const button = document.createElement("button");
     button.textContent = "Prettier";
+    button.type = "button";
     button.classList.add("btn", ...classes);
 
     for (const [key, value] of Object.entries(style)) {
--- a/extension/src/content/index.js
+++ b/extension/src/content/index.js
@@ -71,9 +71,21 @@ function init() {
   }
 
   function setupCommentObserver(observer) {
-    for (const elem of document.querySelectorAll(".timeline-comment-group")) {
+    const JS_DIFF_CONTAINER_CLASS = ".js-diff-progressive-container";
+    const COMMENT_TIMELINE_GROUP_CLASS = ".timeline-comment-group";
+
+    for (const elem of document.querySelectorAll(
+      COMMENT_TIMELINE_GROUP_CLASS
+    )) {
       observer.observe(elem, {
         attributes: true,
+        childList: true,
+        subtree: true
+      });
+    }
+
+    for (const elem of document.querySelectorAll(JS_DIFF_CONTAINER_CLASS)) {
+      observer.observe(elem, {
         childList: true,
         subtree: true
       });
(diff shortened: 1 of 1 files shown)


nickmccurdy:
I tried this myself, but when I press the Prettier button in a pull request inline comment, nothing happens

kaicataldo:
I'm seeing the same behavior as `nickmccurdy` - I don't think this is the fix we want. The behavior this is trying to fix didn't occur in the original implementation. I think it's a regression introduced by https://github.com/prettier/prettier-chrome-extension/commit/befe1fbea6c4ebadb3ad0d271046893ce48fd7a7. Maybe we can try to render the buttons in the same place they were before?

kaicataldo:
Looking at this a bit more, it looks like they are rendered in the same parent node. Maybe GitHub's own JS is binding to this button? I don't think this was happening in the initial implementation because the button was rendered lazily and probably was created after GitHub's scripts ran.

Also, definitely not a bad idea to merge this PR - it just doesn't seem to fix the problem.

kaicataldo:
I'm actually not even seeing our button click handler fire at all at this point. Weird - when I tried this out before we merged it seemed to be working?

Wicked7000:
I'll have a look but when I tested it seemed to be rendering and working okay. From what I've found it seems to be the event propagation that's the problem. (That leads to submitting) so I'll shift that up in both click handlers, otherwise if an error does occur it will submit because the prevent propagation statement hasn't been reached.

kaicataldo:
Any idea why the CI branch check is hanging?

Wicked7000:
> Any idea why the CI branch check is hanging?

Possibly could be this?
https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154

Wicked7000:
> I'm actually not even seeing our button click handler fire at all at this point. Weird - when I tried this out before we merged it seemed to be working?

I've tested this out again locally with and without the type='button' and the click handler fires and prevents it from submitting for me?

kaicataldo:
Sorry, should have clarified. The event handler isn't firing for me for individual line comments (clicking the + button on a line) on the code diff view (in PRs and in individual commits).

Wicked7000:
> Sorry, should have clarified. The event handler isn't firing for me for individual line comments (clicking the + button on a line) on the code diff view (in PRs and in individual commits).

No worries, thanks for letting me know. It seems this was some weirdness around the eventListener being added before the DOM was rendered which meant it didn't attach there is some added elements being observed for when the diff view changes which now updates the listeners which fixes this problem

nickmccurdy:
@kaicataldo @lipis Do you have access to the branch settings? It seems like it's needed to fix the Travis CI build. https://travis-ci.community/t/known-issue-travis-ci-reports-expected-waiting-for-status-to-be-reported-on-the-github-status-api-but-the-status-never-arrives/1154

PR #56, opened 2019-10-23
Implement options page to persist Prettier config

### LINKED ISSUE #25: Config support
https://github.com/prettier/prettier-chrome-extension/pull/23#issuecomment-541471476

nickmccurdy:
Fixes #25

I still need to improve the UI, but feedback is welcome

image


extension/options/src/App.js
```diff
@@ -0,0 +1,140 @@
...
+  tabWidth: 2,
+  trailingComma: "none",
+  useTabs: false
+};
+
+function App() {
+  const [options, setOptions] = useState();
+
+  useEffect(() => {
+    window.chrome.storage.sync.get(defaultOptions, setOptions);
```
kaicataldo: Do we want to scope these settings to a property (`config`, maybe?) in case we want to add other types of data in here later? That way we wouldn't have to make this code backwards compatible in the future.

extension/options/src/App.js
```diff
@@ -0,0 +1,140 @@
...
+  }, [setOptions]);
+
+  useEffect(() => {
+    if (options) {
```
kaicataldo: When would `options` evaluate to falsy here?

extension/options/src/App.js
```diff
@@ -0,0 +1,140 @@
...
+      [name]:
+        type === "checkbox"
+          ? checked
+          : type === "number"
+          ? parseInt(value)
+          : value
+    });
+  }
+
+  return options ? (
```
kaicataldo: When would `options` evaluate to falsy here?

.eslintrc.json
```diff
@@ -2,7 +2,7 @@
   "env": {
     "node": true
   },
-  "plugins": ["prettier"],
+  "plugins": ["prettier", "react"],
```
kaicataldo: Should we turn on the recommended rule set?

extension/options/src/App.js
```diff
@@ -0,0 +1,140 @@
...
+    <>
+      <h1>
+        <a href="https://prettier.io/docs/en/options.html">Prettier Options</a>
```
kaicataldo: Might be good to turn the [`jsx-no-target-blank`]( https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/jsx-no-target-blank.md) rule on.

extension/options/src/App.js
kaicataldo: Seems like we only want to `chrome.storage.sync.get` this data on initial mount, and then we want to `chrome.storage.sync.set` every time time `options` changes (we can rely on this component's internal state for the view after the initial load). What does watching for `setOptions` do here? I'm new to hooks, so might be misunderstanding something :)

nickmccurdy:
> ... I don't think this needs to be a completely separate app. Can we just update the webpack config to output a second bundle? We don't need to compile anything except JSX, I think, so adding and configuring Babel should be fairly straightforward.

I was originally hoping that CRA would make the build setup easier, but it's kind of a hassle to manually run two different build scripts depending on which files change. I'll give it a shot.

extension/options/src/App.js
nickmccurdy: That's a good point, I wasn't thinking about this. It seems like we could store all the options under one object and manually merge them in.

extension/options/src/App.js
nickmccurdy: `setOptions` is asyncronous, so the `options` state isn't updated until the next render, and remains `undefined` on the first call of this `useEffect` hook.

extension/options/src/App.js
nickmccurdy: https://github.com/prettier/prettier-chrome-extension/pull/56#discussion_r337854482

extension/options/src/App.js
nickmccurdy: Is there a rule that bans `target="_blank"` completely (not just when it's insecure)? Screenreaders don't consistently implement support for this, so the extra tab/page can cause accessibility issues.

nickmccurdy: Is there a rule that bans `target="_blank"` completely (not just when it's insecure)? Screenreaders don't consistently implement support for this, so the extra tab/page can cause accessibility issues.

extension/options/src/App.js
nickmccurdy: `eslint-config-react-app` (part of CRA) complains if values are used in `useEffect` without being added to the dependencies array. I'm not sure if it really matters in this case: because the `setOptions` reference is stable, it will only run on the initial mount anyway (as if the dependencies were `[]` or you were using `componentDidMount` in a class component).

extension/options/src/App.js
kaicataldo: Ah, got it! Thanks for the explanation!

extension/options/src/App.js
kaicataldo: Whoops, I thought this had the `target="_blank"` attribute. In this case, might it make sense to add it since this will be any options pop up in Chrome?

extension/options/src/App.js
kaicataldo: Got it, thanks! I think it’s clearer to have it be an empty array, but I’m also new to hooks and defer to your judgment here.

extension/options/src/App.js
nickmccurdy: I'm not sure how links are handled in popups, so I'm trying to test it out. I have this in the manifest (instead of `options_page`) and it's not letting me open the popup, any ideas? ```js
  "options_ui": {
    "page": "options/build/index.html",
    "open_in_tab": false
  },
```

File: extension/options/src/App.js (hunk 1 above)
Comment (nickmccurdy) [reply extension/options/src/App.js]: Whoops I must have been thinking about some previous code, it does actually allow the dependencies here to be `[]` because it knows `setOptiosn` is stable, updated.

**Comment by nickmccurdy** [comment]
Travis seems to be stalling again, I don't even see these commits in the PR builds 😢 
image

File: extension/options/src/App.js (hunk 5 above)
Comment (kaicataldo) [reply extension/options/src/App.js]: Hmmm I'm not sure - I haven't actually made a Chrome Extension settings page before. The popup isn't showing up at all?

**Comment by lipis** [comment]
Seems to be OK now.. or?

File: extension/options/src/App.js (hunk 5 above)
Comment (nickmccurdy) [reply extension/options/src/App.js]: Nope, the button is clicked and nothing shows up, not even an error

File: extension/options/public/index.html [hunk 6]
```diff
@@ -0,0 +1,12 @@
+<!DOCTYPE html>
```
kaicataldo: I think it makes sense to just move this to the root of the `extension` directory (unless we want to create a `public` directory there?).

extension/options/src/App.js
```diff
@@ -0,0 +1,140 @@
+import React, { useEffect, useState } from "react";
```
kaicataldo: Can these files go in `extension/src/options/*`?

webpack.config.js
```diff
@@ -2,8 +2,29 @@ const path = require("path");
...
+    rules: [
+      {
+        test: /\.css$/,
+        use: ["style-loader", "css-loader"]
+      },
+      {
+        exclude: /node_modules/,
+        test: /\.js$/,
+        use: {
+          loader: "babel-loader",
```
kaicataldo: 🙌

extension/options/src/App.js
nickmccurdy: I read your mind, just did this before reading your comment 😎

extension/options/public/index.html
nickmccurdy: It's now in `extension/src/options/index.html`. Since it's only used for options, I don't think it feels right to have it at the top level.

nickmccurdy:
I implemented a custom webpack config and moved the files around a bit. Do we like this setup? Alternatively, how would we feel about moving the contents of `extension` to the root and publishing `dist` instead?

extension/options/public/index.html
kaicataldo: If it's going to live in `src`, can we use [CopyWebpackPlugin](https://webpack.js.org/plugins/copy-webpack-plugin) to copy it into `dist`?

kaicataldo:
I'm not against moving it into the root, but I do think we want to make sure we're only publishing the files needed for the extension, which means we'll need to pick and choose which files to include (`manifest.json/`, `dist/`, `icons/`)

nickmccurdy:
That's what I was thinking of, but I'm not sure of the easiest way to implement it with webpack. I feel like what we have is okay until we start putting larger files in `extension/src` though.

kaicataldo:
I'm happy to do that work with webpack once we land this :)

extension/options/public/index.html
nickmccurdy: That could work, but I feel like it's only worth it if we're just publishing `dist`.

kaicataldo: To be clearer, I don't think we should be shipping anything in `src`. If things don't need to be run through webpack, I would advocate for either using the webpack plugin above or putting them in a directory like `extension/{public,assets}`.

nickmccurdy: I see what you mean. Would you prefer to do this before merging/publishing the options page?

kaicataldo: I’m happy to implement it after.

.eslintrc.json
```diff
@@ -52,7 +49,9 @@
       }
     ],
     "sort-vars": "error",
-    "strict": ["error", "global"]
+    "strict": ["error", "global"],
+    "react-hooks/rules-of-hooks": "error",
+    "react-hooks/exhaustive-deps": "warn"
```
kaicataldo: Can we set this to `"error"` as well? Seems like we would want to block merging for these rules.

nickmccurdy:
This will turn into a normal mergeable PR once we click "Ready for review", but I still have some things I'd like to fix first.

Do you have any preferences on making the UI pretty before or after merging this? I'd be fine either way.

kaicataldo:
Whatever you prefer!

extension/options/src/App.js
nickmccurdy: I still can't get this to work. I'm giving up for now but willing to use a popup if someone else figures it out.

lipis:
let's make it work.. then make it better :)

kaicataldo:
I think this is a good foundation for the settings page, and once we change [this](https://github.com/prettier/prettier-chrome-extension/pull/56#discussion_r337850394) we could merge this and follow up with styling and moving files around!

nickmccurdy:
Yea sorry, still working on that. I ran into some issues and set up devtools to make it easier to troubleshoot, but still need to finish tweaking its configuration.

nickmccurdy:
Now that #59 is merged (which was an interactive rebase of this branch), should I rebase this on top of master?

lipis:
No need  to rebase.. we are squashing the PRs anyways.

nickmccurdy:
I figured out how to scope the `options` property in Chrome's extension storage. I can help out with publishing and UI fixes after this is merged.

lipis:
I couldn't find a way to add more admins to the extension dashboard.. or how could I transfer it to another account so we can enable the automatic publishing.

nickmccurdy:
To be clear, I don't need publishing access for this PR specifically, as the storage is automatically set up with the user's Chrome storage and Google account (if any) due to the `storage` permission in `package.json`.

If you want to set up group publishing see [Set up Group Publishing](https://developer.chrome.com/webstore/publish#set-up-group-publishing). Unfortunately this process is irreversible and charges each new publisher $5, but we could try auto publishing with a single account if you prefer (which I think I would too).

extension/src/content/index.js
```diff
@@ -482,4 +486,4 @@ function init() {
   }
 }
 
-init();
+chrome.storage.sync.get(init);
```
kaicataldo: This isn't necessarily a blocker (we can fix this later if need be), but this only runs when the page is loaded, correct? One use case this might not handle very gracefully is when a GitHub/StackOverflow tab is open and someone goes and changes the settings (they wouldn't go into effect until the page is reloaded).

We could probably fix this by reading from chrome storage each time Prettier formats (again, does not need to be a blocker to land this, just want to make sure it's commented here for posterity).

nickmccurdy: I forgot to implement this, but we should be able to use [`chrome.storage.onChanged.addListener `](https://developer.chrome.com/extensions/storage#event-onChanged)

lipis:
Published in 0.0.4

PR #59, opened 2019-10-29
Improve development with webpack

nickmccurdy:
This is a subset of #56 which only includes improvements to the webpack config so we can review them in isolation and potentially merge them sooner.

Most importantly, `yarn` and `yarn start` should be much faster, and some low priority warnings are hidden.


nickmccurdy:
Thanks for the review. Should we get another before merging? Also please note that this removes the production build from CI, personally I don't think it's useful but I could add it back.

kaicataldo:
LGTM, thanks!

PR #60, opened 2019-10-31
Copy styles and HTML to dist using webpack

kaicataldo:
This updates the webpack config to copy over the `options` markup and styles so that we're not executing code in `src`.


nickmccurdy:
Thanks for this! Looks a lot cleaner.

PR #65, opened 2019-11-01
Connect labels to form inputs properly

nickmccurdy:
This was originally part of [testing-jest](https://github.com/prettier/prettier-chrome-extension/tree/testing-jest) but it's still an important accessibility fix

### DIFF
--- a/extension/src/options/index.js
+++ b/extension/src/options/index.js
@@ -58,95 +58,113 @@ function App() {
         <a href="https://prettier.io/docs/en/options.html">Prettier Options</a>
       </h1>
       <hr />
-      <label>Print width</label>
-      <input
-        type="number"
-        name="printWidth"
-        value={options.printWidth}
-        onChange={handleChange}
-      />
+      <label>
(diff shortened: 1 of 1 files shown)


extension/src/options/index.js
```diff
@@ -58,95 +58,113 @@ function App() {
...
       </h1>
       <hr />
-      <label>Print width</label>
-      <input
-        type="number"
-        name="printWidth"
-        value={options.printWidth}
-        onChange={handleChange}
-      />
+      <label>
```
kaicataldo: For my own learning, how is this better than using the `for` attribute?

nickmccurdy: Moving inputs inside labels automatically associates them in some browsers and accessibility tools. However in theory it's better to do both, if you prefer.

kaicataldo: As long as it works for Chrome, I think we're good :)

kaicataldo:
LGTM, thanks!

PR #66, opened 2019-11-01
Move heading to HTML for popup

nickmccurdy:
This is a different way to fix the layout bug fixed in #61. Instead of adding an extra spacer element, it moves the heading element (which we need to display anyway) to the HTML file. I also replaced `Prettier Options` with just `Options`, as the extension name `Prettier` is already displayed above.

### DIFF
--- a/extension/src/options/index.html
+++ b/extension/src/options/index.html
@@ -2,13 +2,11 @@
 <html lang="en">
   <head>
     <meta charset="utf-8" />
-    <title>Prettier Options</title>
     <link rel="stylesheet" href="options.css" />
   </head>
   <body>
-    <noscript>You need to enable JavaScript to run this app.</noscript>
     <h1>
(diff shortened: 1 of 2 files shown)


kaicataldo:
I initially solved it this way too and I actually decided that this wasn't the right solution. Having the code for the rendered view separated out into various files doesn't feel right to me. I'd love to keep all the view logic in the React component. Happy to reconsider if you feel strongly about this though.

nickmccurdy:
I personally wouldn't consider a static heading to be render logic, I just had it in the JSX because at that point I didn't touch the HTML file for the new React app. That being said, it may make it trickier to replace the whole DOM with a loader or use CSS in JS (such as Styled Components) in the future. It's up to you, but I'd still like to get 902d616d5fe8c2664ff52f9dd722ea11b1f272fe merged at least.

kaicataldo:
Looks like consensus is to do it this way, so let's go with it!

Issue #64, opened 2019-11-01
Fix button styling/location for GitHub


kaicataldo:
The buttons are currently being rendered in such a way that they change the flow of the buttons row. Can we figure out a way for them not to do this?

nickmccurdy:
I'm not sure I understand the issue. In devtools the button seems to be appended at the beginning like I expected. Is it supposed to be a child of the parent div instead?

![image](https://user-images.githubusercontent.com/927220/68055046-2ff7bd80-fcc6-11e9-8366-0d439382b0f2.png)

PR #69, opened 2019-11-02
Fix GitHub button styles

kaicataldo:
Fixes https://github.com/prettier/prettier-chrome-extension/issues/64

The current buttons are styled with `float: left`, which causes some wonkiness with the styles at certain viewport widths. This PR updates the rendering location so that we no longer have to float any elements.

### Editing existing PR opening comment (current):

Screen Shot 2019-11-01 at 11 30 05 PM

### Editing existing PR opening comment (fixed):

Screen Shot 2019-11-01 at 11 31 47 PM

### New comment in PR Files changed view (current):

Screen Shot 2019-11-01 at 11 30 56 PM

### New comment in PR Files changed view (fixed):

Screen Shot 2019-11-01 at 11 31 35 PM

### DIFF
--- a/extension/src/content/index.js
+++ b/extension/src/content/index.js
@@ -10,6 +10,54 @@ import parserTypescript from "prettier/parser-typescript";
 import parserYaml from "prettier/parser-yaml";
 import prettier from "prettier/standalone";
 
+const GITHUB_URL = "https://github.com";
+const GITHUB_VALID_PATHNAMES = /^\/.*\/.*\/(?:pull\/\d+(?:\/?|\/files\/?)$|commit|compare\/.*|issues\/\d+|issues\/new)/u;
+const STACKOVERFLOW_URL = "https://stackoverflow.com";
+const STACKOVERFLOW_VALID_PATHNAMES = /(^\/questions|\/posts\/\d+\/edit)/u;
+
+const PRETTIER_PLUGINS = [
+  parserAngular,
+  parserBabylon,
+  parserFlow,
+  parserGlimmer,
+  parserGraphql,
+  parserHtml,
+  parserMarkdown,
+  parserPostcss,
+  parserTypescript,
+  parserYaml
+];
+const PARSERS_LANG_MAP = {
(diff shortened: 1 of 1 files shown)


extension/src/content/index.js
```diff
@@ -132,33 +131,32 @@ function init() {
...
-      const textArea = findWithClass(prettierBtn, "comment-form-textarea");
-      prettierBtn.addEventListener("click", event => {
-        event.preventDefault();
-        const formattedText = prettier.format(textArea.value, {
-          parser: "markdown",
-          plugins: prettierPlugins,
-          ...storageCache.options
+        const textArea = findWithClass(prettierBtn, "comment-form-textarea");
+
+        prettierBtn.addEventListener("click", event => {
```
kaicataldo: This fixes a memory leak we currently have where this would bind this event listener even when the button existed. This can be demonstrated by going to the Conversation view of a PR, editing an existing comment, and clicking the Prettier button repeatedly.

extension/src/content/index.js
```diff
@@ -10,6 +10,55 @@ import parserTypescript from "prettier/parser-typescript";
 import parserYaml from "prettier/parser-yaml";
 import prettier from "prettier/standalone";
 
+const PRETTIER_PLUGINS = [
```
kaicataldo: Moving all these constants to the top module scope.

extension/src/content/index.js
kaicataldo: This actually isn't the right fix here - it now doesn't bind the Files Changed buttons correctly :( Going to take a different approach.

extension/src/content/index.js
```diff
@@ -10,6 +10,54 @@ import parserTypescript from "prettier/parser-typescript";
 import parserYaml from "prettier/parser-yaml";
 import prettier from "prettier/standalone";
 
+const GITHUB_URL = "https://github.com";
```
kaicataldo: Moving all these constants to the top module scope.

Issue #67, opened 2019-11-02
Move files to root and publish the dist directory directly


nickmccurdy:
https://github.com/prettier/prettier-chrome-extension/pull/56#issuecomment-545709825

kaicataldo:
I think we're currently including the `src` directory in the distributed plugin. We could very easily move everything to the root and use `WebpackCopyPlugin` to copy in whatever other files we need (currently just `icons` and `manifest.json`, I think)?

PR #71, opened 2019-11-02
Move extension files to root directory

kaicataldo:
Fixes https://github.com/prettier/prettier-chrome-extension/issues/67


.eslintrc.json
```diff
@@ -49,21 +49,26 @@
...
-    "react-hooks/rules-of-hooks": "error",
-    "react-hooks/exhaustive-deps": "error"
+    "strict": ["error", "global"]
   },
   "overrides": [
     {
-      "files": ["extension/**/*"],
+      "files": ["src/**/*"],
+      "plugins": ["react-hooks"],
+      "extends": ["plugin:react/recommended"],
```
kaicataldo: Scope the browser-specific configuration to the `src` JS files.

.eslintrc.json
```diff
@@ -49,21 +49,26 @@
...
       "env": {
-        "browser": true
+        "browser": true,
+        "node": false
```
nickmccurdy: Didn't know you could do that, handy! I thought it worked like `globals`.

package.json
```diff
@@ -14,7 +14,7 @@
     "test": "yarn test:other && yarn test:js",
     "test:other": "yarn prettier --list-different",
     "test:js": "eslint --ignore-path .gitignore --ignore-path .prettierignore \"**/*.js\"",
-    "zip": "cd extension && jszip-cli add . -i .DS_Store -o ../extension.zip -f"
+    "zip": "jszip-cli add -i .DS_Store -o extension.zip -f dist/*"
```
nickmccurdy: I'd like to either name this `dist.zip` or move it to `dist` to make the relationship with the other `dist` files more clear.

nickmccurdy:
Looks like we also need to update the readme:
```diff
- - Load unpacked > Select the `extension` directory
+ - Load unpacked > Select the `dist` directory
```

package.json
kaicataldo: Sure, I don't have a strong opinion about this! I think it could make sense to just output everything in `extension/` instead of `dist/`. I'm going to merge this now so we can unblock future PRs, but feel free to make a PR updating this.

kaicataldo: Ah, actually, I'll wait for your approval - didn't realize you requested changes. Please let me know what you think!

nickmccurdy: Are you suggesting that we change `dist` to `extension`? I kinda like that idea since we'll won't need to change the unpackaged extension directory.

kaicataldo: That's right

nickmccurdy: It's up to you if you want to merge this in its current state, I can rename the directory back in another PR

kaicataldo: Okay, sounds good. Let's merge this as is (since @lipis approved the current iteration), and we can always iterate going forward :)

Issue #70, opened 2019-11-02
Bug: Memory leak in GitHub PR Conversation view


kaicataldo:
There's currently a memory leak in the GitHub PR `Conversation` view when editing an existing comment.

#### Steps to reproduce
1. Navigate to an existing PR that you have commented on
2. Edit the comment
3. Click the Prettier button repeatedly

I think this is caused by [these lines of code](https://github.com/prettier/prettier-chrome-extension/blob/master/extension/src/content/index.js#L149-L161), because it is binding the click event listener repeatedly. We can't remove this, though, because it then doesn't bind the button in the `Files changed` view.

nickmccurdy:
If the issue is the same event listener being registered multiple times for the same element, we should be able to fix it by declaring the event listener outside this function scope (if we can access `textArea` from the button's click event). `addEventListener` will not do anything if it has already been called with the same event listener function reference, but it doesn't work with our current code because we create a new function reference on every call (similar issue to creating inline handlers on React components using `React.pure` or `PureComponent` that have expensive render processes).

Issue #53, opened 2019-10-19
Add tests


kaicataldo:
Once the basic functionality has been sorted out, we should split the code into smaller modules and add some tests.

nickmccurdy:
Nice, I was just thinking about this. I feel like since we have so many edge cases with site support (even with only GitHub and Stack Overflow for now), end to end tests with Cypress could be really useful for regression testing.

Wicked7000:
Agreed this would be good to do, although I'm not sure how easy it would be to achieve with Cypress as there are some pretty specific edge cases that I don't know how you could simulate. (At least without just using static web pages)

i.e: Creator of an issue see's a special button when commenting, Owner/Maintainer of a project see's special buttons when commenting on PRs (same with issues) ..etc

I think it would also be good to have some unit tests too though.

nickmccurdy:
It's possible to install a Chrome extension in Cypress, but we may still need to mock the `chrome` APIs and extension URLs. I'm giving this a try, and if it's too difficult to get something working without complex mocks, I might switch to something like Puppeteer or Selenium.

nickmccurdy:
I've tried Cypress and Puppeteer, but they don't seem to support what we need to fully test the extension. Selenium works, but we'd have to authenticate a GitHub account in the tests as the extension only works logged in. I think integration/unit testing would be a better approach, I'm making progress with Jest.

nickmccurdy:
Anyone should be able to start setting up some Jest tests based on #81, but I still have a few in-progress tests that I need to figure out. Please let me know if there's anything you plan on testing so we can coordinate, and only mock/stub what's needed for test environment support so we can keep tests as realistic as possible.

Here are my test plan ideas so far (some of which I've already started implementing):
- Integration test classes and functions in `src/content` with [dom-testing-library](https://github.com/testing-library/dom-testing-library)
- Integration test React components in `src/options` with [react-testing-library](https://github.com/testing-library/react-testing-library)
- Unit/integration test other stuff as needed (I've already tried end to end tests and it's not worth it, Chrome extension support is limited in most frameworks, and even when it's not, it's a pain to deal with GitHub auth)

PR #72, opened 2019-11-02
Refactor content script into modules

kaicataldo:
Here's some preliminary work that separates out the current content script into modules (to make things easier to read and [unit test](https://github.com/prettier/prettier-chrome-extension/issues/53)). This is based on https://github.com/prettier/prettier-chrome-extension/pull/71 and I'll rebase off of `master` once we land that.

I decided to use classes here because there's a good amount of state (including the passed around state of the storage cache) and I think we're going to have to add more. I have to play around with it a bit more, but given the issue reported [here](https://github.com/prettier/prettier-chrome-extension/issues/70), I think we'll want the buttons classes to track their buttons so they can ensure that duplicate bindings aren't occurring.

Alternatively, we could make the storage cache a singleton and just use plain objects or functions with module-scoped state variables if others prefer, but I think this is the easiest to reason about, personally.


nickmccurdy:
FYI you can now tell GitHub that a pull request is a WIP by creating it as a draft: https://github.blog/2019-02-14-introducing-draft-pull-requests/

src/content/storage.js
```diff
@@ -0,0 +1,33 @@
...
+        if (chrome.runtime.lastError) {
+          reject(chrome.runtime.lastError);
+        }
+
+        this._cache = { ...data };
+        resolve();
+      });
+    });
+  }
+}
```
nickmccurdy: I'm glad we're using Promises here, but it would be even nicer to use async/await. Chrome 55+ supports it natively, so we shouldn't need to change our Babel config.

kaicataldo: Do you have a specific pattern in mind? async/await requires a Promise, and so I think we have to wrap this call in a Promise.

src/content/storage.js
```diff
@@ -0,0 +1,33 @@
+export default class Storage {
+  constructor() {
+    this._cache = {};
+    this._initialized = false;
```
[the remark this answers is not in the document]
kaicataldo: I can get behind that :) That being said, this class isn’t a singleton - it just doesn’t re-initialize the state on subsequent `init` calls.

src/content/storage.js
nickmccurdy: I see what you mean now, we're going to need the Promise constructor anyway so async/await wouldn't be useful in this file.

src/content/storage.js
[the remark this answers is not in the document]
nickmccurdy: Is there any case where init is called outside the constructor?

kaicataldo:
Thanks for the review!

> I'm personally not a huge fan of imperative OOP, but the reusability and architecture are greatly improved by these changes.

I'm not married to this - it just made the most sense to me given the changes I want to make next. Here is the [branch](https://github.com/kaicataldo/prettier-chrome-extension/commit/3415e1c7a4148d496b4b76bc033604be46b06bf3) that I'm currently working on. Still needs some fine-tuning, but I think it makes things easier to grok (feel free to disagree 😄).

> Remove the `_` prefix, we don't really need it since we don't have a public API and webpack already wraps each module in a scope (eslint: `no-underscore-dangle: ["error", { "enforceInMethodNames": true }])`)

I think prefixing "private" methods with an underscore is useful because it documents what the intended public API is (not because it actually provides any safety). That being said, if the rest of the team doesn't think this is a good idea, I'm willing to go with consensus.

> Prefer `export default` for primary exports and files with only one export

Do you have an example of which exports you're concerned about? `utils` exporting a single named export is intentional because I imagine it being the module where we can add utility methods going forward.

nickmccurdy:
I'm fine with the OOP and `_` for now. I'd rather have this merged sooner so I can work on tests with less merge conflicts.

> > Prefer `export default` for primary exports and files with only one export
> 
> Do you have an example of which exports you're concerned about? `utils` exporting a single named export is intentional because I imagine it being the module where we can add utility methods going forward.

That's a fair point about `utils`. Other than that, I was referring to `src/content/button.js`, though it may export more functions in the future.

src/content/storage.js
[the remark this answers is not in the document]
kaicataldo: Nope - this is mostly for completeness. I can just delete the code if you don't think it's a good idea.

kaicataldo:
Updated the `button` module's exports!

kaicataldo:
> I'm fine with the OOP and _ for now. I'd rather have this merged sooner so I can work on tests with less merge conflicts.

Sounds good. I think it's great to keep iterating, and am not married to this style of programming. But I agree that having it split into modules is an improvement, regardless!

nickmccurdy:
Ready to merge?

src/content/storage.js
kaicataldo: I updated this to hopefully make it clearer! If we end up writing more wrapper functions around chrome storage methods we can move them into their own file.

kaicataldo:
Mind giving it one more look? Should be ready to go now. Thanks!

PR #73, opened 2019-11-02
Add tests

### LINKED ISSUE #53: Add tests

nickmccurdy:
Fixes #53

- [x] Expose code for testing
- [x] Test basic button injection
- [ ] Create fully-functioning stubs
  - [ ] `chrome.extension.storage.sync` (needs to be custom)
  - [ ] `MutationObserver`
- [ ] Attempt to test with mocked `window.location`
- [ ] Test option syncing behavior
- [ ] Scope ESLint config for tests
- [ ] Consider adding more granular unit tests per class/file/method/function
- [ ] Complete test coverage


package.json
```diff
@@ -51,5 +56,17 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
+  },
+  "babel": {
```
kaicataldo: Can we add this to a babelrc file for consistency?

nickmccurdy: This has the same purpose as `.babelrc` (it's shared between both webpack and jest), but I can extract the config anyway if you prefer.

kaicataldo: We're using .*rc files for our other tooling, and I think it's best to be consistent when possible.

kaicataldo:
This is great! I would advocate for getting the testing infrastructure set up in this PR and landing it so that it frees other people up to start writing tests. What do you think?

nickmccurdy:
I have a bunch of different integration tests for content and options in progress though. What would you consider the testing infrastructure? Just setting up the Jest script and config without any additional libraries?

nickmccurdy:
This got a bit messy with some unrelated test code and configs being combined, so I'm going to split it up into multiple PRs.

PR #75, opened 2019-11-03
Fix memory leak

kaicataldo:
fixes https://github.com/prettier/prettier-chrome-extension/issues/70

This PR simplifies the `MutationObserver` logic and fixes the memory leak described below.

### DIFF
--- a/src/content/github.js
+++ b/src/content/github.js
@@ -1,122 +1,113 @@
+import { findWithClass, isElementVisible } from "./utils";
 import renderButton, { BUTTONS, BUTTONS_TO_SEARCH_FOR } from "./button";
 import { PARSERS } from "./parsers";
-import { findWithClass } from "./utils";
 import prettier from "prettier/standalone";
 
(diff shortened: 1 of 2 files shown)


src/content/github.js
```diff
@@ -1,122 +1,112 @@
...
-        prettierBtn = renderButton(button.parentNode, {
-          append: true,
-          classes: ["prettier-btn"],
-          refNode: button.innerText === SUBMIT_NEW_ISSUE ? button : null,
-          style: { "margin-right": "4px" }
-        });
+      const parentNode = button.parentNode;
+
+      if (
+        !isElementVisible(parentNode) ||
```
kaicataldo: This is what actually fixes the memory leak. Previously, this was rebinding the event handler repeatedly for existing Prettier buttons. This was most likely done because the button in the commit/PR diff views is actually rendered on the page in a non-visible container, and the binding doesn't work when it finally shows up. This differs the creation of the Prettier button until the node it will be rendered into is visible to user (it checks if the element is hidden and if it's in the viewport), and everything now gets bound once!

PR #76, opened 2019-11-03
Fix Create pull request button position

kaicataldo:
The Create pull request button is currently not displaying correctly:
Screen Shot 2019-11-03 at 1 49 36 AM

This is because the button has `float: left;` set, breaking the flow of the buttons.

With this PR, it's fixed to:

Screen Shot 2019-11-03 at 1 48 29 AM

### DIFF
--- a/src/content/button.js
+++ b/src/content/button.js
@@ -1,9 +1,8 @@
-const COMMENT = "Comment";
-const REPLY = "Reply…";
 const CANCEL = "Cancel";
 const CLOSE_ISSUE = " Close issue";
 const CLOSE_PULL_REQUEST = " Close pull request";
-const SUBMIT_PULL_REQUEST = "Create pull request";
+const COMMENT = "Comment";
+const CREATE_PULL_REQUEST = "Create pull request";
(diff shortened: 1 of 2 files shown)


src/content/github.js
```diff
@@ -81,12 +81,29 @@ export default class GitHub {
...
+      const options = {
         append: true,
         classes: ["prettier-btn"],
-        refNode: button.innerText === BUTTONS.SUBMIT_NEW_ISSUE ? button : null,
+        refNode: null,
         style: { "margin-right": "4px" }
-      });
+      };
+
+      // These two buttons have a unique DOM structure, so we need
```
kaicataldo: Trying to simplify and document why we're doing the things we're doing for ease of maintainability. The code has been pretty opaque up to this point.

src/content/button.js
```diff
@@ -1,27 +1,25 @@
-const COMMENT = "Comment";
-const REPLY = "Reply…";
```
kaicataldo: Missed this in my last PR - this is no longer used.

Issue #68, opened 2019-11-02
JSON config import/export


nickmccurdy:
https://github.com/prettier/prettier-chrome-extension/pull/23#issuecomment-541471476

nickmccurdy:
Any preferences on whether this should be part of 1.0?

kaicataldo:
I think this would be nice to implement for v1, but we should finish the current settings panel first.

kaicataldo:
It actually sounds like we had a different idea of what this feature actually is. I wanted to add a text box where users could paste in their local config, and it sounds like the proposal here is to actually import and export from the file system. I don’t think these two features are mutually exclusive - in fact, the import button could populate the text box.

nickmccurdy:
Sorry, I think I was too vague in explaining the UI I was thinking of. Basically, I'd like a JSON input as an alternative to the form, where users can switch back and forth at any time using the same config. This is exactly what [Code Sandbox](https://codesandbox.io/) does, and our form UI is already based on theirs. The thing I like about this is that we don't need to worry about creating separate import/export UI when users can edit the config either way.

> I wanted to add a text box where users could paste in their local config
That's what I intended.

> it sounds like the proposal here is to actually import and export from the file system
That's not what I intended, but since Chrome extensions have more access to the file system, we could potentially add this.

> the import button could populate the text box
I was thinking of this, except that

kaicataldo:
> Sorry, I think I was too vague in explaining the UI I was thinking of. Basically, I'd like a JSON input as an alternative to the form, where users can switch back and forth at any time using the same config. This is exactly what Code Sandbox does, and our form UI is already based on theirs. The thing I like about this is that we don't need to worry about creating separate import/export UI when users can edit the config either way.

Ah okay, thanks for explaining that. I'm not against doing that!

> I wouldn't prioritize it though, as it's easy to copy and paste your config as JSON, and users probably won't need to import/export their config more than once.

Agreed, I don't think this should be a blocker for v1.

PR #77, opened 2019-11-04
Options: add JSON config

kaicataldo:
refs https://github.com/prettier/prettier-chrome-extension/issues/68

Here's an initial implementation of the JSON configuration option. It doesn't have any styles, but all the functionality should be there! Will try to follow up with styles for all the Options next. I _think_ once we complete the styles we'll be ready for v1 - I don't have any other outstanding issues.


src/options/components/App.js
```diff
@@ -0,0 +1,107 @@
...
+  trailingComma: "none",
+  useTabs: false
+};
+
+const defaultOptions = {
+  json: {
+    config: "",
+    enable: false
+  },
+  prettier: defaultPrettierOptions
```
kaicataldo: Sooo after suggesting we scope the options, I realized that `options` is super generic and it gets really confusing to have a nested key called `options`. Just want to make sure this is cool with everyone else - best to fix it now before v1!

src/options/containers/App.js
```diff
@@ -0,0 +1,62 @@
+import React, { useEffect, useState } from "react";
+import App from "../components/App";
+import { promisifiedChromeStorageSyncGet } from "../../shared/chrome";
+
+const defaultOptions = {
+  json: {
+    config: "",
+    enable: false
+  },
+  prettier: {
```
kaicataldo: Sooo after suggesting we scope the options, I realized that options is super generic and it gets really confusing to have a nested key called options. Just want to make sure this is cool with everyone else - best to fix it now before v1!

nickmccurdy:
> Will try to follow up with styles for all the Options next.

I'd like to help out with this, but I wanted to finish the tests first, which could take a while.

> I _think_ once we complete the styles we'll be ready for v1 - I don't have any other outstanding issues.

Would you mind checking [my 1.0 milestone](https://github.com/prettier/prettier-chrome-extension/milestone/1) (which is ordered from high to low priority)? I'd like to know if you think any of these things should wait for a future release like 1.1 or 2.0.

src/content/github.js
```diff
@@ -1,7 +1,8 @@
-import { findWithClass, isElementVisible } from "./utils";
+import { findWithClass, isElementVisible } from "./dom-utils";
```
nickmccurdy: I'd prefer naming the file `domUtils.js` to be consistent with existing files like `stackOverflow.js`.

src/content/options.js
```diff
@@ -0,0 +1,13 @@
+export default function processOptions({ json, prettier: prettierOptions }) {
+  let options = {};
+
+  if (json.enable) {
```
nickmccurdy: I'd prefer to check the presence of `json.config` directly so we don't need an extra variable and this state will have a single source of truth.

src/options/components/JsonConfig.js
```diff
@@ -0,0 +1,87 @@
+import React, { useRef, useState } from "react";
+import PropTypes from "prop-types";
+import parserBabylon from "prettier/parser-babylon";
+import prettier from "prettier/standalone";
+
+const PLACEHOLDER_TEXT = `{
+  "semi": true,
+  "singleQuote": true,
+  "tabWidth": 2
+}`;
```
nickmccurdy: Placeholder text is a good idea, but maybe it would be more useful to display the default options.

src/options/containers/App.js
nickmccurdy: I feel like `prettier` explains less about what they are, especially since there are various API settings for Prettier other than `options`. `prettierOptions` could be more explicit, but personally I think the `prettier` prefix is redundant.

src/options/containers/App.js
```diff
@@ -0,0 +1,56 @@
+import React, { useEffect, useState } from "react";
+import App from "../components/App";
+import { promisifiedChromeStorageSyncGet } from "../../shared/chrome";
+
+const defaultOptions = {
+  json: {
+    config: "",
+    enable: false
+  },
```
nickmccurdy: If we remove this from `defaultOptions`, we can store the JSON directly in `defaultOptions.options` and use the same value's truthiness to check its existence.

src/content/options.js
kaicataldo: I don’t think this would be a great user experience, because it means you have to clear out the JSON config manually even if you uncheck the option to use the JSON config.

src/options/containers/App.js
kaicataldo: Not sure I follow - mind explaining what you mean?

src/options/containers/App.js
kaicataldo: `prettierOptions` sounds good to me!

src/options/components/JsonConfig.js
kaicataldo: Sure, that sounds good to me

src/content/options.js
nickmccurdy: To be clear, I mean we would also remove the checkbox option.

kaicataldo: Can you explain why you think that would be better, from a user experience standpoint?

kaicataldo:
> Separating components into separate files is a great idea, but I prefer to avoid components directories, as it's more useful to organize files by domain/usage than type.

Do you have an alternative for how you think it should be organized right now? Or are you thinking of keeping a flat directory structure in `src/options/` right now? In a project this small, I don't think there is much to gain or lose by moving files around, so I don't think it's worth bikeshedding too much on it. If you have another way you think works better, happy to move things around.

src/options/containers/App.js
kaicataldo: As a follow up, I don't think we want to store this config alongside the other Prettier options, because we pass those directly to Prettier when we format code.

nickmccurdy:
> > Separating components into separate files is a great idea, but I prefer to avoid components directories, as it's more useful to organize files by domain/usage than type.
> 
> Do you have an alternative for how you think it should be organized right now? Or are you thinking of keeping a flat directory structure in `src/options/` right now? In a project this small, I don't think there is much to gain or lose by moving files around, so I don't think it's worth bikeshedding too much on it. If you have another way you think works better, happy to move things around.

I was thinking of a flat directory structure to keep things simple for now, like what we have in `src/content`.

nickmccurdy:
I explained more about my original UI idea in https://github.com/prettier/prettier-chrome-extension/issues/68#issuecomment-549489261, though we don't have to follow it if you don't want to.

src/content/options.js
nickmccurdy: I changed my mind, we don't need to store this config twice. We can use `options`/`prettierOptions`/whatever to store the parsed config for both the form and JSON editor. If you'd like we can also have a synced option to toggle whether the form or JSON editor is used, otherwise it would default to one of them and keep the state locally in React.

kaicataldo:
> I was thinking of a flat directory structure to keep things simple for now, like what we have in src/content.

Sounds fine to me.

> I explained more about my original UI idea in #68 (comment), though we don't have to follow it if you don't want to.

Thanks for explaining that. I'm definitely not opposed to doing it that way, and here are some thoughts/discussion points around it:

1. If we do it this way we can get rid of storing the JSON string at all, and go back to having one `options` (`prettierOptions`?) key, and just derive the text box value by running `JSON.stringify()` on the options in storage.
1. My biggest question is around validation/supported options. I guess we could keep a whitelist of supported options in the version of Prettier we're running (I don't think there's a programmatic API to get this information, but I don't think these change very often so it's probably not a big deal), and display an error message if there are properties in the config that aren't supported. I want to avoid a situation where we're storing arbitrary data in Chrome storage.

nickmccurdy:
> 1. If we do it this way we can get rid of storing the JSON string at all, and go back to having one `options` (`prettierOptions`?) key, and just derive the text box value by running `JSON.stringify()` on the options in storage.

Exactly. We could even run Prettier when loading the settings so the formatting is consistent regardless of how it was created.

> 2. My biggest question is around validation/supported options. I guess we could keep a whitelist of supported options in the version of Prettier we're running (I don't think there's a programmatic API to get this information, but I don't think these change very often so it's probably not a big deal), and display an error message if there are properties in the config that aren't supported. I want to avoid a situation where we're storing arbitrary data in Chrome storage.

Adding validation is a good idea, but figuring out edge cases of how to run it could be tricky. What if a user pastes in an invalid config and then switches to the form (which can't be displayed properly)? When do we validate and how?

I can't seem to find an API for the config options either. Could we use the keys of `defaultOptions` instead? Those are the only options supported by the form UI anyway, so we wouldn't have to worry about removing any to make them compatible. We could also add

kaicataldo:
> Adding validation is a good idea, but figuring out edge cases of how to run it could be tricky. What if a user pastes in an invalid config and then switches to the form (which can't be displayed properly)? When do we validate and how?

Right - this is my biggest concern and why I went with what seemed like the simpler approach of giving control to the user (if it's invalid, it's on the user to make it valid). That being said, I think there's an easy fix. Currently, there's a "Save" button that we can utilize for validation (it currently validates whether it's valid JSON or not, but we include checking the keys for valid whitelisted properties as well). If it's not valid, it shows an error and doesn't update the configuration. We should also be able to add some nice error messaging (`doubleQuote is not a valid Prettier configuration option`, or `singleQuote must be a boolean`). I'll work on implementing this.

nickmccurdy:
Thanks. I don't mean to make this too complex. Free to split up PRs if it's useful and let me know if you need help.

kaicataldo:
Sorry for the delay - finally was able to get back to this. I regenerated the lockfile because I ended up having merge conflicts and was worried about editing it manually 😬

Please let me know what you think! The validation is strict, but I think it has to be so that we don't allow users to save arbitrary data in Chrome Storage.

src/options/options.js
```diff
@@ -0,0 +1,39 @@
...
+    {}
+  );
+  const errors = [];
+
+  for (const [key, val] of Object.entries(options)) {
+    if (!Object.hasOwnProperty.call(validationMap, key)) {
+      errors.push(`${key} is an invalid option`);
+    }
+
+    if (typeof val !== validationMap[key]) {
```
kaicataldo: All of the current options are all a consistent, single type. This might have to be updated in the future if we support any options that can be multiple types.

src/options/components/JsonConfig.js
kaicataldo: We don't need a placeholder any more because we're always displaying what the user has configured.

kaicataldo:
@nickmccurdy Friendly ping - does the changes ease your concerns?

src/options/JsonConfig.js
```diff
@@ -0,0 +1,85 @@
...
+
+const SAVED_TIMEOUT = 500;
+const STRINGIFY_SPACING = 2;
+
+export default function JsonConfig({ errors, options, setErrors, setOptions }) {
+  const textareaEl = useRef(null);
+  const [textAreaVal, setTextAreaVal] = useState(
+    JSON.stringify(options),
+    null,
+    STRINGIFY_SPACING
```
nickmccurdy: Looks like the closing parent is in the wrong place: `null` and `STRINGIFY_SPACING` should be arguments for `JSON.stringify` rather than `useState`

src/options/JsonConfig.js
```diff
@@ -0,0 +1,85 @@
...
+  useEffect(() => {
+    try {
+      setTextAreaVal(
+        prettier.format(JSON.stringify(options), {
+          parser: "json",
+          plugins: [parserBabylon],
+          ...options
+        })
+      );
+    } catch {}
```
nickmccurdy: If this throws when the text field is invalid, I'd like to show (or at least log) an error.

src/options/JsonConfig.js
```diff
@@ -0,0 +1,85 @@
...
+    setDisplaySaved(true);
+    window.setTimeout(() => {
+      setDisplaySaved(false);
+    }, SAVED_TIMEOUT);
+  }
+
+  return (
+    <>
+      {errors.map((err, idx) => (
+        <p key={idx}>Error: {err}</p>
```
nickmccurdy: Could we use the error name or message as a unique key? Index keys should be avoided in most cases.

src/options/options.js
nickmccurdy: Prettier doesn't seem to have any [options](https://prettier.io/docs/en/options.html) like this, so it should be fine.

src/shared/chrome.js
```diff
@@ -0,0 +1,11 @@
...
+  return new Promise((resolve, reject) => {
+    chrome.storage.sync.get(data => {
+      if (chrome.runtime.lastError) {
+        reject(chrome.runtime.lastError);
+      }
+
+      resolve(data);
+    });
+  });
+}
```
nickmccurdy: Maybe we could write a whole high-level Promise-based API on top of the `chrome` API so we don't have to deal with stubbing as much in tests, doesn't have to be in the PR though

kaicataldo: Sure, that seems reasonable

src/options/JsonConfig.js
kaicataldo: Ha, good catch

src/options/App.js
```diff
@@ -0,0 +1,58 @@
...
+    </>
+  );
+}
+
+App.propTypes = {
+  errors: PropTypes.array,
+  options: PropTypes.object,
+  setErrors: PropTypes.func,
+  setOptions: PropTypes.func
+};
```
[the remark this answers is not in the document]
kaicataldo: I’m definitely not against this - we could follow up with this

src/shared/chrome.js
nickmccurdy: I usually use TypeScript now but I can give this a shit

src/options/App.js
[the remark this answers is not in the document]
nickmccurdy: I usually use TypeScript now but I can give this a shot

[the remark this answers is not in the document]
kaicataldo: I’ve never been a big fan of using the `PropTypes` package. Definitely prefer using a real type system (though I’m not necessarily advocating for that for this repo).

src/shared/chrome.js
kaicataldo: If we do that, it might be nice for one of us to just create a separate package that this project consumes - I’m sure others would use it :)

src/options/JsonConfig.js
kaicataldo: What would you want the error to say? I’m not sure how useful an error here would actually be to the user. That being said, I understand the fear of swallowing errors silently. I guess we could log something in the console?

src/options/App.js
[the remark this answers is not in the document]
nickmccurdy: I'd be ok with switching to TypeScript, I avoided it at first because I wasn't sure how many other Prettier contributors would have experience

src/options/JsonConfig.js
nickmccurdy: Could be a simple "Error: Invalid JSON at ..." message in the form

src/shared/chrome.js
nickmccurdy: I originally suggested [webextension-polyfill](https://github.com/mozilla/webextension-polyfill) for Firefox support, but it could be useful for this as it brings Firefox's Promise-based extension API to Chrome

src/options/App.js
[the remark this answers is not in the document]
kaicataldo: Yeah, I'm not sure it's worth it for a small project like this - feels like it raises the bar for contribution unnecessarily.

src/options/JsonConfig.js
kaicataldo: That should be covered already though, since we JSON parse it. This was more for safety than anything else - didn't want an internal error in Prettier to bubble up and break the rest of the extension.

kaicataldo: Can we table this for now and see if it's really an issue? Like I said, this was more so that some error out of our control couldn't bring the whole extension down.

kaicataldo: I'm not opposed to adding something here, but I do want to get this merged so we can continue towards v1!

kaicataldo:
This is ready for re-review. Would love to land this!

src/options/JsonConfig.js
nickmccurdy: Yea that's fine with me, I resolved this. I think I misread the code at first, it's not that unsafe.

src/shared/chrome.js
kaicataldo: Oh yeah, that sounds like a good idea!

nickmccurdy: Started in #99

kaicataldo:
Thanks! We can talk about this separately, but how would you foresee autosave working without constantly showing “malformed JSON” errors?

nickmccurdy:
Good question, we'd probably have to display errors when invalid and only update the saved config when valid

kaicataldo:
I initially actually did implement it that and thought the UX was frustrating. If we can come up with a way to do that that doesn’t have guaranteed errors in your face as you type, I’m definitely down for exploring that!

PR #88, opened 2019-11-07
Split up configs

nickmccurdy:
- ~~I removed the env configs since they don't seem to be in use. If we want to use them I can enable the `no-undef` rule.~~
- Both ESLint and Babel configs are split up by directory.
- Babel is technically only needed for `src/options` but has an empty config elsewhere in case we add options outside this directory in the future.
- We can add overrides for things like tests as needed.

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,9 +1,9 @@
 {
+  "env": {
+    "es2020": true
+  },
   "plugins": ["prettier"],
   "extends": ["prettier"],
-  "parserOptions": {
-    "ecmaVersion": 2020
-  },
   "rules": {
     "curly": "error",
     "dot-notation": "error",
@@ -20,6 +20,7 @@
       }
     ],
     "no-shadow": "error",
+    "no-undef": "error",
     "no-unneeded-ternary": "error",
     "no-unused-expressions": "error",
     "no-unused-vars": [
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,12 +1,8 @@
 {
-  "env": {
-    "node": true
-  },
   "plugins": ["prettier"],
   "extends": ["prettier"],
   "parserOptions": {
-    "ecmaVersion": 2020,
-    "sourceType": "module"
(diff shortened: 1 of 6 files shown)


kaicataldo:
Can we just use `overrides` so we can have a single config file for each tool? Both Babel and ESLint support this feature.

nickmccurdy:
I was thinking that splitting them up in directories would make them easier to find and move around.

kaicataldo:
I personally tend to avoid relying on the merging behavior of configuration files by directory since glob overrides were introduced, since you have to go between files and merge the contents of nested configs mentally. But if others feel strongly about it, this definitely works too.

.eslintrc.json
```diff
@@ -1,12 +1,8 @@
 {
-  "env": {
-    "node": true
```
kaicataldo: Don’t we want this for `webpack.config.js`?

src/.eslintrc.json
```diff
@@ -0,0 +1,5 @@
+{
+  "parserOptions": {
```
kaicataldo: I think we’re missing the browser env here?

kaicataldo:
That’s right, they add in globals. I’m a bit confused why you would want to remove them?

nickmccurdy:
I can add them back if you prefer. I removed them because I was under the impression that the globals were only used in `no-undef` (which we aren't using) and were therefore unnecessary. I'd personally prefer to use `no-undef` with the globals though.

kaicataldo:
Ah, I see - I didn’t realize `no-undef` was disabled. I think we should definitely turn that rule on and add the globals back!

nickmccurdy:
Yea sorry for the confusion! Will do, I've been missing that rule anyway.

.eslintrc.json
nickmccurdy: Added to `webpack.config.js`

src/.eslintrc.json
nickmccurdy: Fixed.

PR #99, opened 2019-11-16
Use webextension-polyfill's Promise-based API

nickmccurdy:
Trying out my idea from https://github.com/prettier/prettier-chrome-extension/pull/77#discussion_r347072207

Note that there are multiple ways to bundle this. I'm currently just bundling a copy which each entry (making both entries 35KB larger in production). Alternatively we could code split our shared modules (might help with tree shaking) or copy the original package's scripts straight into the `extension` directory (would require adding to `manifest.json`).


kaicataldo:
I’m fine with focusing on feature-completeness over saving bytes right now, particularly since this will install once and then load from the file system. Definitely great to try to optimize later though!

nickmccurdy:
Technically it's in the bundle twice now, but I'm not sure I really care as it's just an extra 35K (Prettier is pretty huge in comparison)

nickmccurdy:
Is there anything you’d like me to change?

kaicataldo:
> Technically it's in the bundle twice now, but I'm not sure I really care as it's just an extra 35K (Prettier is pretty huge in comparison)

Yeah, and since they're two scripts that are executing in completely different contexts, it seems fine to me.

PR #130, opened 2020-01-06
Add LeetCode support

chrissantamaria:
Hi all! I've been using LeetCode (popular coding interview prep site) recently and really wished I could add Prettier integration to their site's editor. After talking with @nickmccurdy, he introduced me to this awesome project and I realized I could just contribute to this rather than make my own extension.

LeetCode uses a library called [CodeMirror](https://codemirror.net/) for their integrated editor (specifically, [react-codemirror2](https://github.com/scniro/react-codemirror2)). Though it's fairly tricky to get / set the editor's content by inspecting the DOM alone, the editor's CodeMirror instance is exposed as a property of the editor DOM element (`.CodeMirror`). This makes it super easy to call CodeMirror methods:

```js
const editor = document.querySelector(".CodeMirror");
// Get the editor's content
editor.CodeMirror.doc.getValue();
// Set the editor's content
editor.CodeMirror.doc.setValue('some code');
```

Unfortunately, this property is only accessible within a page context - as extension content scripts are sandboxed, this CodeMirror instance can't be used in the same way. To get around this, I've created some helper methods that are injected into the page context to get / set the editor's content (see `_addClientListeners`).

Right now, this implementation works, though there are some things I'm not crazy about:

- I'm currently injecting the client script by creating a script element, setting its `textContent` through a template literal, and injecting that into the DOM. [This StackOverflow post](https://stackoverflow.com/questions/9515704/insert-code-into-the-page-context-using-a-content-script/9517879#9517879) nicely summarizes some other possible ways to run code within a page context - I'm open to other methods.
- Event listener hell. Right now the flow is "GetEditorContent" -> "FormatEditorContent" -> "SetEditorContent", but this isn't super easy to follow. I'm not sure what a better way to handle client-extension communication would be.
- Right now this implementation assumes the editor is set to JavaScript. I think this is the only language we could support right now with the current parser setup (unless we add community plugins, but that's another conversation), but we should probably remove the Prettier button when the editor language is set to something other than JavaScript.
- I'm not super familiar with MutationObserver and mostly just copied the StackOverflow implementation. I'm pretty sure there's a more specific query to observe than `body` and I might not need the `childList` and `subtree` options - open to optimizations there.
- I'm manually grabbing class names for button styling, but they seem to be auto-generated classes that might not stay constant with new LeetCode changes. Ideally, I'd like to either apply those styles manually or find a more consistent way of grabbing them from LeetCode.

I'm leaving this as a draft PR as I think it still needs some more work before being merged. Very open to any thoughts you guys may have!

### DIFF
--- a/manifest.json
+++ b/manifest.json
@@ -16,7 +16,11 @@
   "permissions": ["storage"],
   "content_scripts": [
     {
-      "matches": ["https://stackoverflow.com/*", "https://github.com/*"],
+      "matches": [
+        "https://stackoverflow.com/*",
+        "https://github.com/*",
+        "https://leetcode.com/*"
+      ],
       "js": ["content.js"],
       "run_at": "document_end"
     }
--- a/src/content/extension.js
+++ b/src/content/extension.js
@@ -1,10 +1,13 @@
 import GitHub from "./github";
+import LeetCode from "./leetCode";
 import StackOverflow from "./stackOverflow";
 import Storage from "./storage";
 
 const GITHUB_URL = "https://github.com";
 const STACKOVERFLOW_URL = "https://stackoverflow.com";
 const STACKOVERFLOW_VALID_PATHNAMES = /(^\/questions|\/posts\/\d+\/edit)/u;
+const LEETCODE_URL = "https://leetcode.com";
+const LEETCODE_VALID_PATHNAMES = /(^\/problems)/u;
 
 export default class Extension {
   constructor() {
@@ -24,6 +27,13 @@ export default class Extension {
       STACKOVERFLOW_VALID_PATHNAMES.test(window.location.pathname)
     ) {
       this._buttons = new StackOverflow(this._storage);
+    }
+
+    if (
+      window.location.origin === LEETCODE_URL &&
+      LEETCODE_VALID_PATHNAMES.test(window.location.pathname)
+    ) {
+      this._buttons = new LeetCode(this._storage);
     }
   }
 }
--- a/src/content/leetCode.js
+++ b/src/content/leetCode.js
@@ -11,8 +11,8 @@ export default class LeetCode {
   _init() {
     this._addClientListeners();
 
-    this._createButton();
-    const pageObserver = new MutationObserver(() => this._createButton());
+    this._handleChange();
+    const pageObserver = new MutationObserver(() => this._handleChange());
     pageObserver.observe(document.querySelector("body"), {
       childList: true,
       subtree: true
@@ -52,6 +52,19 @@ export default class LeetCode {
     script.remove();
   }
 
+  _handleChange() {
+    const languageEl = document.querySelector('[data-cy="lang-select"]');
+    if (!languageEl) {
+      return;
+    }
+
+    if (languageEl.innerText !== "JavaScript") {
+      this._removeButton();
+    } else {
+      this._createButton();
+    }
+  }
+
   _createButton() {
     const buttonRowEl = document.querySelector('[class^="btns"]');
     if (!buttonRowEl) {
@@ -82,6 +95,13 @@ export default class LeetCode {
       // Initiating format event chain
       document.dispatchEvent(new Event("GetEditorContent"));
     });
+  }
+
+  _removeButton() {
+    const button = document.querySelector(".prettier-btn");
+    if (button) {
+      button.remove();
+    }
   }
 
   _addExtensionListener() {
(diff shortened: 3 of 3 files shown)


nickmccurdy:
> * I'm currently injecting the client script by creating a script element, setting its `textContent` through a template literal, and injecting that into the DOM. [This StackOverflow post](https://stackoverflow.com/questions/9515704/insert-code-into-the-page-context-using-a-content-script/9517879#9517879) nicely summarizes some other possible ways to run code within a page context - I'm open to other methods.
> * Event listener hell. Right now the flow is "GetEditorContent" -> "FormatEditorContent" -> "SetEditorContent", but this isn't super easy to follow. I'm not sure what a better way to handle client-extension communication would be.

I'm not sure I fully understand the alternatives. It would be nice to have some simpler function calls, but if the other ones make it too hard to share code I'm fine with these events.
> * Right now this implementation assumes the editor is set to JavaScript. I think this is the only language we could support right now with the current parser setup (unless we add community plugins, but that's another conversation), but we should probably remove the Prettier button when the editor language is set to something other than JavaScript.

What other languages are supported besides javascript? Here are the parsers we support: https://github.com/prettier/prettier-chrome-extension/blob/master/src/content/parsers.js

> * I'm manually grabbing class names for button styling, but they seem to be auto-generated classes that might not stay constant with new LeetCode changes. Ideally, I'd like to either apply those styles manually or find a more consistent way of grabbing them from LeetCode.

Personally this seems like the most important improvement to me, as we wouldn't be sure when LeetCode could regenerate their styles and break the DOM manipulation logic. It would be nice if we can find more stable CSS classes or HTML tags to locate the same elements.

chrissantamaria:
> I'm not sure I fully understand the alternatives. It would be nice to have some simpler function calls, but if the other ones make it too hard to share code I'm fine with these events.

The only other reasonable alternative I see would be the first method described in the linked post: putting the page context script in a separate file and adding it to `web_accessible_resources`. This might just make the approach even more confusing so for now I'm okay leaving it as is.

As for event listeners, the main issue is just following the chain of events. Perhaps it's worth even briefly describing the flow in a comment block at the top of the file?

> What other languages are supported besides javascript? Here are the parsers we support: /src/content/parsers.js@master

Tons! C++, Java, Python, C, C#, Ruby, Swift, Go, Scala, Kotlin, Rust and PHP. Obviously supporting all of these isn't a priority, but I think it's still a good idea to hide the button when on a non-supported language. I'll try to implement this soon.

> Personally this seems like the most important improvement to me, as we wouldn't be sure when LeetCode could regenerate their styles and break the DOM manipulation logic. It would be nice if we can find more stable CSS classes or HTML tags to locate the same elements.

It looks like some elements actually have `data-cy` properties set on them including the languages select dropdown (`data-cy="lang-select"`) - I'll look into this more and see if I can dynamically grab the select's classes and apply them to the Prettier button.

chrissantamaria:
I managed to replace to hardcoded class names using the `data-cy` property mentioned above. I also made a `_removeButton` method to run when the language is set to something other than JavaScript.

My only lingering concern now is the frequency of the MutationObserver calls. Like I said in my initial comment, I'm almost certain there's a more specific querySelector than `body`, but I'm not entirely sure what that would be. Right now, either `_createButton` or `_removeButton` are called on every single DOM update after the initial load which, though it hasn't impacted my performance noticeably, is probably not a great idea.

chrissantamaria:
I've gone ahead and marked this as ready for review - I'll leave it up to you guys to decide if the MutationObserver is fine for now. @lipis @kaicataldo if you guys have any thoughts on this let me know.

nickmccurdy:
I'm fine with the mutation observer usage personally, though I've barely used them so take my opinion with a grain of salt, might want another opinion

chrissantamaria:
Just checking on the status of this - not sure if we want to keep thinking about the mutation observer or just go ahead and merge

chrissantamaria:
Like I said, from my (limited) understanding of mutation observers, I think it's firing on every single DOM update. This isn't _terrible_ since it's only updating the Prettier button when necessary, but it's still not ideal.

lipis:
Let's merge and improve later.

lipis:
How can I transfer the app to somebody else to do the releases?

nickmccurdy:
This is working on GitHub Actions (our new CI). Do we need anything else before merging?

chrissantamaria:
Nope, I think we're okay based on the thread above.

nickmccurdy:
Thanks, and sorry for the delay, I wasn't sure about the status

PR #187, opened 2020-03-27
Add GitHub Actions workflow for Node.js CI

nickmccurdy:


nickmccurdy:
Travis is still reporting a check even though I disabled its webhook. Should we try deleting the webhook or disabling it from the website?

nickmccurdy:
I just noticed they have their own readme badges, should I add one? Personally I could go without it, as the repository's home page still shows the status in the commit check status.

kaicataldo:
I don't have a strong opinion about this :)

PR #242, opened 2020-06-10
Update Options UI

sibiraj-s:
Previous: 

Screenshot 2020-06-08 at 1 51 45 PM

Screenshot 2020-06-10 at 1 41 27 PM
 
Now: 

Screenshot 2020-06-10 at 1 28 54 PM

Screenshot 2020-06-10 at 1 28 59 PM

Made few changes for now, we can make more changes in separate iterations.


sibiraj-s:
eslint rules like `sort-imports` and `sort-keys` should be loosened up a little, the dev env is sort of too strict.

sort import rule appears to have some bug it seems. It always throws error no matter what i do.

lipis:
Yes if it's not doing it automatically it's hard to find the right sort order by hand :) Feel free to remove that rule: https://github.com/prettier/prettier-chrome-extension/blob/master/.eslintrc.json#L39

sibiraj-s:
done @lipis

lipis:
Thanks!

PR #245, opened 2020-06-11
Add support for firefox browser

### LINKED ISSUE #46: Cross browser support with WebExtensions
Supporting multiple browsers used to involve writing a different codebase for each set of APIs, but now most major browsers support the WebExtensions standard. I’m guessing the next most requested browser would be Firefox, but we could do a poll to figure out what browsers Prettier users would like to use and research from there.

sibiraj-s:
- Add support for firefox browser

extension will be built into
- `extension/chrome`
- `extension/firefox`

zip file will be built into
- `artifcats/prettier-chrome-v{version}.zip`
- `artifcats/prettier-firefox-v{version}.zip`

To start chrome dev

```
yarn build:chrome:dev
```
To start firefox dev

```
- yarn build:firefox:dev
- yarn lauch:firefox // opens a firefox instance with extension loaded
```

closes #46


sibiraj-s:
I have made only minimal changes to the code to create the build.

Running yarn build will create two artifacts
- chrome.zip
- web-ext-artifacts/prettier-{version}.zip (for Firefox)


I was thinking about moving the build into separate scripts that would be more manageable 🤔

.eslintrc.json
```diff
@@ -44,15 +44,20 @@
         "natural": true
       }
     ],
-    "sort-vars": "error",
-    "strict": ["error", "global"]
```
sibiraj-s, line 48: remove since es6 by default in strict mode and webpack also automatically adds it to the output

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 extension
-extension.zip
+DS_Store
```
lipis, line 3:
```suggestion
.DS_Store
```

.eslintrc.json
kaicataldo: This is intentional, and is used for configuration files that are still written in CommonJS.

webpack.config.js
```diff
@@ -1,77 +1,82 @@
-/* eslint-env node */
-"use strict";
```
kaicataldo, line 2: I would prefer to keep this in strict mode since this isn't an ES module.

.gitignore
nickmccurdy: .DS_Store should not be ignored in project gitignores, only your global gitignore (see [Configuring ignored files for all repositories on your computer](https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer))

.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 extension
-extension.zip
+.DS_Store
```
nickmccurdy, line 3: .DS_Store should not be ignored in project gitignores, only your global gitignore (see [Configuring ignored files for all repositories on your computer](https://help.github.com/en/github/using-git/ignoring-files#configuring-ignored-files-for-all-repositories-on-your-computer))

kaicataldo: Why is that?

nickmccurdy: Because these files are system. There are many different files that could be created by systems, editors, etc. It's better for projects to only ignore files that all users will have regardless of their system (such as node_modules in Node projects) as we won't be able to predict what machines/editors/etc. are being used by contributors.

kaicataldo: Fair enough! I don't think it does much harm, but totally get your point :+1:

nickmccurdy:
This seems to work well on Firefox when installed temporarily (without signing), thank you! Does this set up what we need to sign the extension for Firefox, or is there more work to be done for that?

sibiraj-s:
See https://extensionworkshop.com/documentation/publish/signing-and-distribution-overview/#distributing-your-addon. When published to through the AMO developer hub, it is not required to sign the extension.

nickmccurdy:
Great. I'm fine with merging this if everyone else is.

sibiraj-s:
Pushed a small commit to merge manifest files efficiently. will come in handy while using using different config in prod vs development env.

sibiraj-s:
@nickmccurdy yarn.lock is not deleted

nickmccurdy:
Ah sorry, you're right, must have been looking at the wrong part of the diff

nickmccurdy:
@sibiraj-s Do you have this published on AMO yet? If so, please PR to link to it. If not, would you like help?

sibiraj-s:
@nickmccurdy I haven't published it. I thought, it will be better if some member publishes the official extension. I could publish if the team agrees.

kaicataldo:
I can take that on. I also need to publish the latest version for Chrome.

nickmccurdy:
Thanks! Let me know if you need help though, might have some time in the next few days

sibiraj-s:
cool.

just a thought. may be do we need to rename the repo to something generic like `prettier-browser-extension` or `prettier-web-extension`.  🤔

nickmccurdy:
Maybe just `prettier-extension` or `prettier-extensions`?

sibiraj-s:
`prettier-extension` that sounds better too, but it feels too generic?

kaicataldo:
I like `prettier-browser-extension`!

kaicataldo:
I don't think so!

sibiraj-s:
@kaicataldo Hi. just a ping on the release, can we publish the latest changes in the master to chrome and also make it available in Firefox too. those changes are really useful.

Is there any blockers or anyway I could help?

kaicataldo:
Thanks for the offer to help. I'm currently waiting on https://github.com/prettier/prettier-browser-extension/pull/353 to be reviewed. I didn't want to publish to Firefox and then immediately publish a completely different version.

If you'd like to review that PR, I'd appreciate that!

sibiraj-s:
Sure. I can do a review tomorrow.
________________________________

PR #283, opened 2020-07-14
Migrate to native dependabot

sibiraj-s:
Lets Migrate to native dependabot?

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,6 +1,6 @@
 version: 2
 updates:
-    # Enable version updates for npm
+  # Enable version updates for npm
   - package-ecosystem: "npm"
     # Look for `package.json` and `lock` files in the `root` directory
     directory: "/"
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,9 @@
+version: 2
+updates:
+    # Enable version updates for npm
+  - package-ecosystem: "npm"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,9 @@
+version: 2
+updates:
+  # Enable version updates for npm
+  - package-ecosystem: "npm"
```
nickmccurdy, line 4: I assume this refers to the registry, not the package manager, right? If not we may need to change it to Yarn

nickmccurdy:
What's the difference with what we already have? Would we have to authorize the third party Dependabot app in the organization?

.github/dependabot.yml
sibiraj-s: > I assume this refers to the registry, not the package manager.

correct.  it refers to ecosystem, not the manager. package manager is automatically detected.

sibiraj-s:
this is what we have now https://github.com/marketplace/dependabot-preview.
 
`Dependabot is moving natively into GitHub!` blog link https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/.

> Would we have to authorize the third party Dependabot app in the organization?

since dependabot is now native to github, just merging should do. Prettier main repo is already using this

May be the old app `dependabot-preview` has to be disabled (i am not completely sure about this.)

Issue #21, opened 2019-10-11
Additional ways to execute formatting


nickmccurdy:
These are some ideas based on what I think is useful though I'm not sure we'd want all of them:

- Keyboard shortcut ([prettier-js-extension](https://github.com/krawaller/prettier-js-extension) uses 
 `Ctrl + Shift + F`)
- Context menu
- Button action (clicking on the extension icon in Chrome's extension toolbar)
- Automatic formatting (could work like the VSCode extension and allow auto formatting on any supported file, possible with a debounce/throttle if we hit performance bottlenecks)

kaicataldo:
The only one I have reservations about is the automatic formatting. Otherwise, I like this idea a lot! I also think it would be nice to add a feature where, if text is selected, it will only format that text.

nickmccurdy:
> The only one I have reservations about is the automatic formatting.
Out of curiosity, why don't you like about this? How would you feel about adding it with a toggle that's off by default?

> I also think it would be nice to add a feature where, if text is selected, it will only format that text.
As far as I'm aware Prettier requires formatted text to be valid as an entire file, so this would mean that only some selections could be formatted properly. Personally I think it's easier to only allow formatting entire inputs, since it's closer to Prettier's built in functionality.

kaicataldo:
> Out of curiosity, what don't you like about it? How would you feel about adding it with a toggle that's off by default?

Having a toggle sounds good! I don't think it's a good idea to modify text as users are typing, since characters can move and make it really unwieldy.

> As far as I'm aware Prettier requires formatted text to be valid as an entire file, so this would mean that only some selections could be formatted properly. Personally I think it's easier to only allow formatting entire inputs, since it's closer to Prettier's built in functionality. Otherwise we would have to explain to the user why selections are invalid and implement extra error handling.

That's right. What I like about this functionality is that blocks of code that aren't in markdown can be formatted. That being said, I don't think this should be a high priority feature.

Issue #20, opened 2019-10-11
Generic text box support


nickmccurdy:
Other extensions like [prettier-js-extension](https://github.com/krawaller/prettier-js-extension) allow any text box to be used for formatting, which would help make this extension more useful with less code changes. I still think some of the functionality and UI specific to GitHub and Stack Overflow could be useful though.

kaicataldo:
Out of curiosity, how do we want to handle detecting the parser? Just naively assume it's Markdown? I wonder if the we could have a dropdown list when clicking the Prettier button in the toolbar that says `Format as...` and then has a list of langs we're supporting?

nickmccurdy:
Good question, maybe we could use a language detection library and fall back to Markdown when it's unknown or looks like snippets from multiple languages.

kaicataldo:
Ohhh I like that idea :)

nickmccurdy:
Some resources for language detection:
- https://prettier.io/docs/en/api.html#prettiergetsupportinfo-version
- https://highlightjs.readthedocs.io/en/latest/api.html#highlightauto-value-languagesubset
- https://www.npmjs.com/package/language-map
- https://www.npmjs.com/package/language-detect

PR #290, opened 2020-07-19
Firefox Only - Add Option to format via context menu

sibiraj-s:
Partially addresses #20 , #21 

In chrome it seems the element is not accessible on context menu events. hence added only to Firefox for now. 

Screenshot 2020-07-19 at 7 24 09 PM


kaicataldo:
Related discussion: https://github.com/prettier/prettier-chrome-extension/issues/322

nickmccurdy:
Thanks for this. I think `contex-menu.js` should be `context-menu.js`. Also have you been able to confirm that this works in our supported sites on Firefox, and the existing non-menu functionality still works in Chrome?

sibiraj-s:
> I think contex-menu.js should be context-menu.js

AutoComplete 😅 . updated now.

> Also have you been able to confirm that this works in our supported sites on Firefox, and the existing non-menu functionality still works in Chrome?

Yes. I verified it. There is no changes to existing code and the context-menu script is loaded only in Firefox.

src/background/context-menu.js
```diff
@@ -0,0 +1,38 @@
...
+  });
+  await browser.tabs.sendMessage(tab.id, "prettierFormat");
+};
+
+const CLASS_NAME = "__PRETTIER_EXTENSION_FORMAT_ELEMENET__";
+
+const menuItem = {
+  contexts: ["editable"],
+  id: menuItemId,
+  title: "Format with prettier",
```
kaicataldo, line 20:
```suggestion
  title: "Format with Prettier",
```

kaicataldo:
Thanks for the PR! I'd like to implement https://github.com/prettier/prettier-browser-extension/issues/322 before we add any other functionality to the extension. I'm definitely not against considering this in the future once that's completed, but I think it's best to start with the new approach and see what we feel is missing.

kaicataldo:
#322 has been merged. We should re-evaluate this in light of the new behavior.

sibiraj-s:
Thanks @kaicataldo . 

RightClick to format is a general option available mostly everywhere. I think it would be nice to include the feature. May be with the current changes. I may be even make it work across the browsers.

sibiraj-s:
Yes It is a nice option. Till we figure out language detection. It will be very handy.

I do remember giving a thought on this when I was working with this feature. I think there is a limitation. An Extension can have only one context menu option. Adding more than one will automatically group them into sub menus. 

My memory is little vague on that. I will check and update back

sibiraj-s:
Yes. It is not possible to render more than one context menu item

> You can create as many context menu items as you need, but if more than one from your extension is visible at once, Google Chrome automatically collapses them into a single parent menu.

https://developer.chrome.com/docs/extensions/reference/contextMenus/#usage

sibiraj-s:
@kaicataldo I have updated the context logic to align with existing one. 

One known issue. ContextMenu will show up on all editable fields like input, contentEditable. but the format will work only in textarea. Need to handle that.

sibiraj-s:
@kaicataldo can we merge this?

PR #301, opened 2020-07-27
Enable chrome styled input elements

sibiraj-s:
- Enable chrome styled input elements, so that it will look consistent across platform in chromium based browsers
- disable spell check in json config editor

**Before:**

Screenshot 2020-07-27 at 11 04 21 AM

**After**

Screenshot 2020-07-27 at 11 04 42 AM

### DIFF
--- a/src/manifest.json
+++ b/src/manifest.json
@@ -11,7 +11,8 @@
   },
   "options_ui": {
     "page": "options.html",
-    "open_in_tab": false
+    "open_in_tab": false,
+    "chrome_style": true
   },
   "permissions": ["storage"],
   "content_scripts": [
--- a/src/options/JsonConfig.js
+++ b/src/options/JsonConfig.js
@@ -67,6 +67,7 @@ export default function JsonConfig({ errors, options, setErrors, setOptions }) {
         value={textAreaVal}
         ref={textareaEl}
         onChange={handleChange}
+        spellCheck="false"
       />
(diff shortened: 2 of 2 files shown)


kaicataldo:
Thanks for contributing!

nickmccurdy:
What you're seeing in the "Before" screenshot is actually the [updated form controls for Chromium](https://blog.chromium.org/2020/03/updates-to-form-controls-and-focus.html). As far as I'm aware Chromium will not use native form UI in either scenario, so it's always a consistent cross platform look. However, now our settings page uses the old form UI while Chromium itself and other extensions use the new form UI.

I'll revert the form UI change, but we still appreciate your help and the `spellCheck` prop change is useful.

sibiraj-s:
> As far as I'm aware Chromium will not use native form UI in either scenario, so it's always a consistent cross platform look

I was wrong here.

It doesn't give consistency with platform. it enables consistency with other elements in the settings page. 

> If this is included and set to true, then your document will get one or more extra stylesheets that will help make it look consistent with the browser's UI and with other extensions that use the browser_style property

Reference See https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/user_interface/Browser_styles.

may be we can add browser_style property too 🤔

nickmccurdy:
Isn't that the default now? If I'm understanding the article correctly.

PR #305, opened 2020-08-02
Support stackexchange sites other than stackoverflow.

### LINKED ISSUE #90: Support Stack Exchange sites other than Stack Overflow
Stack Overflow is in the Stack Exchange network of Q&A sites. They all have similar formatting and implementations, so supporting the other sites should be relatively easy.

In order to do this, we'll need a whitelist of all Stack Exchange domains (not all of them use subdomains of https://stackexchange.com/). I recommend separating out the GitHub and Stack Exchange support into separate content scripts so we don't have to duplicate this whitelist into the JS source.

## Lists of Stack Exchange sites
- https://stackexchange.com/sites
- https://meta.stackexchange.com/a/81383/232041

sibiraj-sr:
Support stackexchange sites

Fixes #90

### DIFF
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
     "fix:js": "yarn test:js --fix",
     "prettier": "prettier . --write",
     "build:chrome:dev": "PLATFORM=chrome node scripts/start.js",
-    "build:firefox:dev": "PLATFORM=chrome node scripts/start.js",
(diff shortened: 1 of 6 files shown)


sibiraj-sr:
I think all the major sites are included 🤔 .

PR #306, opened 2020-08-02
Support stackexchange sites other than stackoverflow

### LINKED ISSUE #90: Support Stack Exchange sites other than Stack Overflow

sibiraj-s:
Support stackexchange sites

Fixes #90


sibiraj-s:
I think most of the sites are included 🤔 . let me know if anything is missing. I will add.

lipis (approved):

src/content/extension.js
```diff
@@ -1,39 +1,24 @@
...
-      this._buttons = new StackOverflow(this._storage);
-    }
-
-    if (
-      window.location.origin === LEETCODE_URL &&
-      LEETCODE_VALID_PATHNAMES.test(window.location.pathname)
-    ) {
-      this._buttons = new LeetCode(this._storage);
+    for (const Extension of extensions) {
+      if (typeof Extension.test === "function" && Extension.test()) {
```
kaicataldo, line 18: It doesn't look like any of the extensions are missing `#test()`. If that's true, I'd prefer not to have this guard so that it actually throws if one of the Extensions doesn't have this static method rather than silently not work.

sibiraj-s: agreed. removed.

kaicataldo:
Thanks for contributing!

Issue #258, opened 2020-06-23
Add screenshot(s) to `README.md` showing usage


with-heart:
I discovered this extension today and was able to figure out how it works, but screenshots would make that much more obvious.

Anita-ihuman:
I will like to take on this issue

kaicataldo:
@Anita-ihuman That would be great, thanks!

Anita-ihuman:
Please review my PR@nickmccurdy

PR #316, opened 2020-08-19
Added instructions for setting options .

Anita-ihuman:


nickmccurdy:
Hi @Anita-ihuman, did you close this by accident? It seems like useful information. If you forgot to commit a fix for #258 here, you can reopen the PR and push to the same branch to update it with your new changes.

PR #320, opened 2020-08-19
Simplify CI setup

kaicataldo:
There's currently a lot of indirection in our dev/Ci setup (built up over time). This PR  aims to split tasks into easier to understand groupings, separate concerns, and make our scripts more self-explanatory/explicit. I also split out linting from formatting, because, after nearly 5 years of working on linters, I think formatting should be a separate concern.


lipis (changes requested):
Should update the `README.md` as well..

chrissantamaria:
Looks like there's still two "Test in Chrome locally" sections. Probably best to either combine the commands into one section like you did (though maybe add `yarn launch:firefox` as well) or leave all Firefox-related instructions in the second section (and fix the title).

CI stuff looks good! I think the scripts are much more streamlined now.

kaicataldo:
@chrissantamaria Thanks! Fixed.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,26 @@
+name: CI
+on:
+  push:
+    branches:
+      - master
+  pull_request:
+    branches:
+      - master
```
nickmccurdy, line 8: Does this mean pull requests to or from master? We're rarely going to have a pull request from master.

package.json
```diff
@@ -61,11 +59,8 @@
     "webpack-cli": "3.3.12"
   },
   "lint-staged": {
-    "*.{js}": [
-      "eslint --ignore-path .gitignore --ignore-path .prettierignore --fix"
-    ],
```
nickmccurdy, line 66: I think we should keep ESLint enabled. It's useful to run on change files as the user may not have an editor integration, especially now that `yarn test` isn't running ESLint.

package.json
```diff
@@ -11,16 +11,14 @@
...
+    "dev:firefox": "PLATFORM=firefox node scripts/start.js",
     "launch:firefox": "web-ext run",
-    "test": "yarn test:other && yarn test:js && jest --passWithNoTests",
-    "test:other": "prettier . --check",
-    "test:js": "eslint --ignore-path .gitignore --ignore-path .prettierignore \"**/*.js\""
+    "lint": "eslint --ignore-path .gitignore \"**/*.js\"",
+    "format": "prettier . --write",
+    "format:check": "prettier . --check",
+    "test": "jest --passWithNoTests",
+    "ci": "yarn format:check && yarn test"
```
nickmccurdy, line 21: I don't think we need this `ci` script, please refer to my comment about enabling ESLint in `lint-staged`.

package.json
kaicataldo: I'm personally really not a fan of linting on commit, since I commit often and and tend to amend commits. I think it's a good point about not running the linter for `yarn test`, though. Maybe can add the linting script back to `test` and then have a separate script for CI?

.github/workflows/ci.yml
kaicataldo: Should be to `master`

package.json
sibiraj-s: Optionally we can run lint/tests before pushing to remote by adding it to pre-push step using husky.

kaicataldo: Addressed these comments. Please re-review.

kaicataldo:
Would love to land this since it's just an internal chore. Please re-review!

sibiraj-s (approved):

PR #323, opened 2020-08-19
Added instructions for setting options

Anita-ihuman:


kaicataldo:
Thanks for contributing! It looks like CI is failing because Prettier isn't being run on the README.md file. This should happen automatically when you make a commit. Did you run `yarn` to install dependencies locally? This could also be fixed by running [`yarn fix`](https://github.com/prettier/prettier-chrome-extension/blob/master/package.json#L14) locally.

README.md
```diff
@@ -66,7 +66,15 @@ yarn fix
 ```
 yarn test
 ```
+### Setting up
```
kaicataldo, line 69: This is a bit confusing because the rest of this file is geared towards developers working on this project. If we want to add usage documentation, I think we should either create a new, separate section of the readme or create a new file. This information would ideally be on a real site - I wonder if we could add documentation on the Prettier website once this becomes stable?

Anita-ihuman: Actually I think adding documentation should be considered and well also contributor guide in the repositories on how to go about Contributing to Prettier. I don't think I have been it anywhere

Anita-ihuman: By creating another section, do I use wiki pages? for that or I should just leave it for it to be included in the docs?

kaicataldo: That's a really great point about missing a `CONTRIBUTING.md` document. We should definitely add this!

I think a Wiki Page would be great! @lipis @nickmccurdy What are your thoughts?

Anita-ihuman:
@kaicataldo please can you review

README.md
nickmccurdy: The standard on GitHub is putting development information in CONTRIBUTING.md, it will even notify users about the file if they open a pull request for the first time. It seems like our development and help sections should be moved to contributing, while installation and this new setting up section should be in the readme. While we're at it, it would be nice to show some basic build instructions for Firefox users until we have it officially published.

package.json
```diff
@@ -72,5 +72,7 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
-  }
+  },
+  "main": "index.js",
+  "author": "anita-ihuman <charlesanita403@gmail.com>"
```
nickmccurdy, line 77: If you'd like to add your name to authors I think it would be best to use an array of all collaborators

package.json
```diff
@@ -72,5 +72,7 @@
     "hooks": {
       "pre-commit": "lint-staged"
     }
-  }
+  },
+  "main": "index.js",
```
nickmccurdy, line 76: main shouldn't be needed since this isn't a package
```suggestion
```

Anita-ihuman: Uhmmm
Thing I wasn’t sure how to  run the yarn fix
So I just did a few trials and the code went through 
So I am guessing I should just take out the  main right?

nickmccurdy: Yes, this shouldn't affect script execution.

Anita-ihuman: Yes, I do need help, please.
for a clearer understanding of how this works

package.json
kaicataldo: I'd like to avoid having an author field. They add extra maintenance burden for little benefit, since we can already see the authors in the commit list and repo stats.

package.json
kaicataldo: Yeah, we don't need to have add this (though I suppose there's little harm in doing so).

To run `yarn fix`, you'll need to follow these steps:

1. [Install yarn](https://classic.yarnpkg.com/en/docs/install)
2. Navigate to the root directory of this project (where `package.json` is located)
3. Run `yarn` on the command line (or `yarn install`) to install the dependencies
4. Run `yarn fix` on the command line

package.json
nickmccurdy: Ah right, I think this is more useful for published packages

Anita-ihuman:
Can you please review this PR again? @nickmccurdy

nickmccurdy (commented):
Thanks, the `package.json` looks good now. I still think it would help to split up parts of the readme into `CONTRIBUTING.md` as I mentioned above, as there's a lot of information for developers that ordinary extension users don't need. Let me know if you need help with that.

PR #324, opened 2020-08-20
Grab manifest version from package.json

chrissantamaria:
Hi all 👋 this PR is just a quality of life change to prevent having to maintain two versions in `package.json` and `src/manifest.json`. merge-json-webpack-plugin exposes a handy property `beforeEmit` that I used to put the `package.json` version into the final manifest.


kaicataldo:
Thanks, this is really helpful!

PR #328, opened 2020-08-24
Update prettier to v2.1

### LINKED ISSUE #203: Unable to load unpacked extension after development build
## Issue

Attempting to load the latest version of the extension after a development build (`yarn start`) gives the following error:

However, loading the extension folder after a production `yarn build` (ignoring the final compressed zip) works as expected.

I ran a quick `git bisect` and found that 7cbbc20 was the first commit to introduce this. I'm still trying to narrow down exactly what is causing this - seems like a fairly innocent commit as far as webpack goes. Even reverting the changes to `webpack.config.js` still causes the issue. @nickmccurdy, any thoughts?

## Steps to reproduce

1. Clone the repo
2. `yarn`
3. `yarn start`
4. [Load the unpacked extension in Chrome](https://github.com/prettier/prettier-chrome-extension#test-in-chrome-locally)

## Environment

- macOS Catalina 10.15.3
- Chrome 80.0.3987.163
- Node 12.16.2
- Yarn 1.21.1

sibiraj-s:
Update prettier to v2.1.0

closes #203

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "classnames": "2.2.6",
-    "prettier": "2.0.5",
+    "prettier": "2.1.0",
     "prop-types": "15.7.2",
     "react": "16.13.1",
     "react-dom": "16.13.1",
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "classnames": "2.2.6",
-    "prettier": "2.1.0",
+    "prettier": "2.1.1",
     "prop-types": "15.7.2",
     "react": "16.13.1",
     "react-dom": "16.13.1",
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,6 +1,5 @@
 "use strict";
 
-const TerserPlugin = require("terser-webpack-plugin"); // included as a dependency of webpack
 const CopyPlugin = require("copy-webpack-plugin");
 const { CleanWebpackPlugin } = require("clean-webpack-plugin");
 const HTMLPlugin = require("html-webpack-plugin");
@@ -45,18 +44,6 @@ module.exports = ({ outDir, env }) => {
             "sass-loader",
           ],
         },
-      ],
-    },
-    optimization: {
-      minimizer: [
-        new TerserPlugin({
-          terserOptions: {
-            extractComments: false,
-
-            // https://github.com/webpack-contrib/terser-webpack-plugin/issues/107
-            output: { ascii_only: true },
-          },
-        }),
       ],
     },
     output: {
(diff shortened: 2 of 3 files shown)


nickmccurdy:
Thanks, this does fix the issue in development! I'm noticing the production build is a megabyte larger, but is 10 times faster than on master. Thoughts on if this is acceptable or if we should stick to similar terser options since production builds are less common?

# master build
```
yarn run v1.22.4
$ PLATFORM=chrome node scripts/build.js
Hash: 9394f6adbd19593723ae
Version: webpack 4.44.1
Time: 132588ms
Built at: 08/25/2020 3:12:48 AM
     Asset       Size  Chunks                    Chunk Names
content.js   7.47 MiB       0  [emitted]  [big]  content
REDACTED
✨  Done in 135.63s.
```

# this build
```
yarn run v1.22.4
$ PLATFORM=chrome node scripts/build.js
Hash: a519ca1d8ac56bf92b9a
Version: webpack 4.44.1
Time: 10306ms
Built at: 08/25/2020 3:10:11 AM
     Asset       Size  Chunks                    Chunk Names
content.js   6.33 MiB       0  [emitted]  [big]  content
REDACTED
✨  Done in 13.84s.
```

sibiraj-s:
Also, in coming days. if possible. we can move this to background scripts rather than content-scripts. so that we don't need to load them in main pages.

sibiraj-s:
Actually this build is a smaller than the master build 🤔 

**Master**: content.js   7.47 MiB
**Current Branch**: content.js   6.33 MiB

nickmccurdy:
@kaicataldo I meant changing the webpack config, not avoiding the upgrade

kaicataldo:
I believe I added the `TerserPlugin` explicitly so that we could add `output: { ascii_only: true },`. Since this PR fixes the bug in dev, relying on Webpack's defaults for production minification sounds good to me.

nickmccurdy:
> Actually this build is a smaller than the master build 🤔

I'm getting similar results, I must have mixed them up, sorry! I'll merge this.

> Also, in coming days. if possible. we can move this to background scripts rather than content-scripts. so that we don't need to load them in main pages.

I'm not sure how this would work or what the advantage would be, but feel free to make an issue or pull request if you're still interested in it.

sibiraj-s:
> I'm not sure how this would work or what the advantage would be

some advantages are.

- background scripts are not loaded in users site.
- background scripts doesn't have strict utf checks.
- background scripts are loaded only once when the extension is installed. (content is loaded from the storage so loading only once doesn't give any particular advantage. just adding a point)

> but feel free to make an issue or pull request if you're still interested in it.

Sure. I will evaluate if this is really helpful and propose something.

PR #329, opened 2020-08-25
Disable chrome styled input elements

nickmccurdy:
This reverts the UI changes from #301 ([explanation](https://github.com/prettier/prettier-chrome-extension/pull/301#issuecomment-679779109)) without reverting the disabling of `spellCheck`.


kaicataldo:
Makes sense, thanks!

PR #337, opened 2020-09-03
Fix loading Chrome extension from production build

nickmccurdy:
#328 has fixed the development build in Chrome (tracked in #203), but it seems like it also broke the production build (as we still need the custom terser options for Prettier 2).

@sibiraj-s @kaicataldo Sorry I didn't notice this until after merging #328, but it's still helpful for fixing the development build.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -1,5 +1,6 @@
 "use strict";
 
+const TerserPlugin = require("terser-webpack-plugin"); // included as a dependency of webpack
 const CopyPlugin = require("copy-webpack-plugin");
 const { CleanWebpackPlugin } = require("clean-webpack-plugin");
 const HTMLPlugin = require("html-webpack-plugin");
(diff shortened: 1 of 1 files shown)


nickmccurdy:
In retrospect I think it would be a good idea for us to have both builds end to end tested in CI. I tried to use `web-ext run --source-dir extension/chrome --target chromium` and it does show a popup with the error, but unfortunately doesn't change the status code so we can fail the build. Would appreciate help figuring this out.

kaicataldo (approved):

fisker:
@nickmccurdy 

Does this mean this PR https://github.com/prettier/prettier/pull/8534 in Prettier is not good enough? 
Prettier already add `ascii_only: true` for all files now, [here](https://github.com/prettier/prettier/pull/8983/files#diff-42eca0e47b0b26d08270b8ba15b08850R189).

You probably should keep the `ascii_only: true`, just let you guys know about it.

nickmccurdy:
I'm not sure, Prettier 2.1.2 has the same issue and it seems like this fix hasn't been released yet.

PR #338, opened 2020-09-03
Install web-ext (used in scripts)

nickmccurdy:
It looks like we forgot to add this to our `devDependencies` when adding the `launch:firefox` script.


sibiraj-s:
Yeah. I forgot to push that change earlier. I added it here in #290. but it should have been a separate PR.

sibiraj-s (approved):

PR #352, opened 2020-10-01
chore: use overrides instead of config merging for ESLint config

kaicataldo:
Small refactor of our config so that it's consolidated in one file.


.eslintrc.json
```diff
@@ -1,9 +1,14 @@
 {
-  "env": {
-    "es2020": true
-  },
+  "root": true,
```
kaicataldo, line 2: We were missing this

.eslintrc.json
```diff
@@ -49,15 +54,27 @@
...
-        "jest": true
+        "browser": true,
+        "node": false
+      },
+      "parserOptions": {
+        "sourceType": "module"
       }
     },
     {
-      "files": ["scripts/**/*.js", "webpack.config.js"],
```
kaicataldo, line 58: We can just lint all JS files

.eslintrc.json
```diff
@@ -1,9 +1,14 @@
 {
-  "env": {
-    "es2020": true
-  },
+  "root": true,
   "plugins": ["prettier"],
   "extends": ["prettier"],
+  "env": {
+    "es2020": true,
+    "node": true
```
kaicataldo, line 7: We can just lint all JS files

package.json
```diff
@@ -14,7 +14,7 @@
     "dev:chrome": "PLATFORM=chrome node scripts/start.js",
     "dev:firefox": "PLATFORM=firefox node scripts/start.js",
     "launch:firefox": "web-ext run",
-    "lint": "eslint --ignore-path .gitignore \"**/*.js\"",
+    "lint": "eslint --ignore-path .gitignore .",
```
kaicataldo, line 17: We can just lint all JS files

kaicataldo:
Going to go ahead and merge this since it seems like everyone else on the project is currently unavailable.

PR #353, opened 2020-10-04
Rewrite extension to listen for keybinding

kaicataldo:
fixes https://github.com/prettier/prettier-browser-extension/issues/322

Sorry for the delay! Here's a first pass at a rewrite. It now listens for a keybinding (defaulting to `Ctrl + Shift + ,` but this is customizable via the native browser extension APIs) and formats the text. As you can see, it allows us to delete _a lot_ of code. I didn't want to create any premature abstractions, but we can always refactor in the future :) I initially wanted to keep all the non-DOM based logic in the background script (to keep the injected script as lightweight as possible), but I was worried about the potential of hitting the size limit for message passing payloads. Maybe this isn't a real concern? If we decide this isn't a concern, we can make separate concerns a lot more cleanly here. This should also improve the ease with which we can test things!

Please let me know what you think. Once this merges, I'll publish both for Chrome and Firefox.

Side note: I disabled the `build` check because it appears to be broken and is failing on all PRs. Anyone know what this actually does?


src/background/index.js
```diff
@@ -0,0 +1,11 @@
+/* globals chrome */
+
+chrome.commands.onCommand.addListener((command) => {
```
kaicataldo, line 3: Background script listens for keybindings and uses message passing to notify the client of the action

src/content/extension.js
```diff
@@ -1,24 +1,261 @@
...
-import LeetCode from "./leetCode";
-import StackOverflow from "./stackOverflow";
+/* globals chrome */
 
+import prettier from "prettier/standalone";
 import Storage from "./storage";
+import { PARSERS, PARSERS_LANG_MAP } from "./parsers";
 
-const extensions = [GitHub, StackOverflow, LeetCode];
+const STACKEXCHANGE_SITES = [
```
kaicataldo, line 7: Inlined all this for now but we can always create more abstractions in the future if this becomes more complex

src/background/index.js
```diff
@@ -0,0 +1,11 @@
+/* globals chrome */
```
kaicataldo, line 1: We can move this to the ESLint config once https://github.com/prettier/prettier-browser-extension/pull/352 is merged.

kaicataldo:
@lipis @nickmccurdy Friendly ping :)

kaicataldo:
@lipis @nickmccurdy One more friendly ping :)

kaicataldo:
@lipis @nickmccurdy I'm happy to take the lead and just merge stuff, just don't want to rogue and start changing stuff y'all aren't supportive of. At this point, this PR is holding up any future work, so would be great to get this merged.

kaicataldo:
Let me know if you're too busy and I can just get to work :)

kaicataldo:
I'm going to wait another day or two and then go ahead and merge this so we can move the project forward. Hope everyone is doing okay!

src/content/extension.js
```diff
@@ -1,24 +1,261 @@
...
+    );
+  }
+
+  _format(inputEl) {
+    if (this._isStackOverflow()) {
+      this._formatStackOverflow(inputEl);
+    }
+
+    this._formatDefault(inputEl);
+  }
```
sibiraj-s, line 61:
```suggestion
  _format(inputEl) {
   const options = this._storage.get().prettierOptions;
  
    if (this._isStackOverflow()) {
      this._formatStackOverflow(inputEl, options);
    }

    this._formatDefault(inputEl, options);
  }
```

instead of getting it multiple times during execution (in _formatStackOverflow)

src/content/extension.js
```diff
@@ -1,24 +1,261 @@
...
+    });
+  }
+
+  _isStackOverflow() {
+    const origin = window.location.origin;
+
+    return (
+      (STACKEXCHANGE_SITES.some((url) => url === origin) ||
+        STACKEXCHANGE_URL_REGEX.test(origin)) &&
+      STACKEXCHANGE_VALID_PATHNAMES.test(window.location.pathname)
```
sibiraj-s, line 51: `STACKEXCHANGE_VALID_PATHNAMES ` 

Do we need this? Since we do generic text box formatting? It should be able to format anything inside stackoverflow?

src/content/extension.js
```diff
@@ -1,24 +1,261 @@
...
+
+  _bindListeners() {
+    chrome.runtime.onMessage.addListener((message, sender, sendResponse) => {
+      if (message.action !== "runPrettierFormat") {
+        return;
+      }
+
+      const inputEl = document.activeElement;
+
+      if (!inputEl) {
```
sibiraj-s, line 36: document.activeElement will always exist. we need filter it for supported input elements. like only textarea.

for now we can do.

```suggestion
      if (inputEl?.tagName === 'TEXTAREA') {
```

I don't think text input boxes needs to be supported as no one is going to use it for code.

src/background/index.js
sibiraj-s: this works on both Firefox and chrome. should we consider using `webextension-polyfill` instead for consistency?

sibiraj-s (changes requested):
Few small things

sibiraj-s:
should we enable this for all URL's in content-scripts?

```
"matches": [
  "<all_urls>"
],
```

src/content/extension.js
kaicataldo: Good idea +1:

lipis (dismissed):
Go rogue..!

nickmccurdy (approved):
The code looks good so far.

nickmccurdy:
Something to keep in mind for the future is that only using keyboard shortcuts could limit our ability to format blocks that aren't Markdown. Can we figure out a way to automatically detect the language or are we going to need to add some sort of language selector popup eventually?

kaicataldo:
@nickmccurdy I believe that's a limitation of the current implementation as well. Language detection seems like a doable thing (try and parse with all the available parsers until one sticks!), and seems like an interesting thing to explore 😄

nickmccurdy:
I mean they have different UI constraints for how we could select languages. For the current inline button approach, we could add a dropdown to manually select icons. But if no UI is added to the page itself, we'll either have to make it automatic or add a new UI, like a popup or dropdown menu of languages.

kaicataldo:
I'm going to try to get back to this this week. Would love to land this and get it published :)

src/content/extension.js
kaicataldo: I think we still do. StackOverflow uses its own special flavor of markdown, unfortunately...

src/background/index.js
kaicataldo: Awesome. Let's do this in a future PR (I need to test in Firefox still anyway!)

kaicataldo:
Thanks for the reviews, folks! I'm going to go ahead and merge this so we can start iterating from there.

PR #378, opened 2020-11-20
Switch to Prettier's ESM build

nickmccurdy:
https://prettier.io/blog/2020/11/20/2.2.0.html#esm-standalone-bundles-8983httpsgithubcomprettierprettierpull8983-by-monchihttpsgithubcommonchi-fiskerhttpsgithubcomfisker

Now we're importing from Prettier's new ESM standalone build instead of its CJS standalone build

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "classnames": "2.2.6",
-    "prettier": "2.2.0",
+    "prettier": "2.2.1",
     "prop-types": "15.7.2",
     "react": "16.14.0",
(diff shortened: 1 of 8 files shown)


sibiraj-s (dismissed):

package.json
```diff
@@ -22,7 +22,7 @@
   },
   "dependencies": {
     "classnames": "2.2.6",
-    "prettier": "2.1.1",
+    "prettier": "2.2.0",
```
sibiraj-s, line 25: latest is now v2.2.1

PR #398, opened 2021-02-01
Remove .dependabot.yml file

kaicataldo:
I don't think this is particularly helpful or useful, especially since the majority of these PRs sit unmerged. Since we have Dependabot security notifications turned on, we'll still be notified of upgrades necessary for security reasons.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Setup Node.js
-        uses: actions/setup-node@v1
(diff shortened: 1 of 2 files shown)


sibiraj-s:
Agreed. 👍

.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
       - name: Setup Node.js
         uses: actions/setup-node@v1
```
sibiraj-s, line 16:
```suggestion
        uses: actions/setup-node@v2
```

kaicataldo:
Thanks for the review!

